PR #1, opened 2021-10-04
go mod tidy and fix generator

deejross:
This fixes the conflict between go vet and controller-gen.


api/v1alpha1/commons.go
```diff
@@ -10,7 +10,7 @@ import (
 type KubeAuthConfiguration struct {
 	// ServiceAccount is the service account used for the kube auth authentication
 	// +kubebuilder:validation:Required
-	// +kubebuilder:default={Name:&#34;default&#34;}
+	// +kubebuilder:default={Name:default}
```
raffaelespazzoli, line 13: this was intentional and needed because that how you do default for a structure.

api/v1alpha1/databasesecretenginerole_types.go
```diff
@@ -38,12 +38,12 @@ type DatabaseSecretEngineRoleSpec struct {
 
 	// DeafulTTL Specifies the TTL for the leases associated with this role. Accepts time suffixed strings ("1h") or an integer number of seconds. Defaults to system/engine default TTL time.
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default=0s
+	// +kubebuilder:default=0
```
raffaelespazzoli, line 41: this was intentional, are you sure that it should be `0`.

api/v1alpha1/databasesecretenginerole_types.go
```diff
@@ -38,12 +38,12 @@ type DatabaseSecretEngineRoleSpec struct {
...
 	DefaultTTL metav1.Duration `json:"defaultTTL,omitempty"`
 
 	// MaxTTL Specifies the maximum TTL for the leases associated with this role. Accepts time suffixed strings ("1h") or an integer number of seconds. Defaults to system/mount default TTL time; this value is allowed to be less than the mount max TTL (or, if not set, the system max TTL), but it is not allowed to be longer. See also The TTL General Case.
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default=0s
+	// +kubebuilder:default=0
```
raffaelespazzoli, line 46: this was intentional, are you sure that it should be `0`.

raffaelespazzoli (changes requested):

api/v1alpha1/databasesecretenginerole_types.go
deejross: The generator was complaining about an unexpected 's'

api/v1alpha1/commons.go
deejross: The generator was complaining about unexpected characters.

api/v1alpha1/databasesecretenginerole_types.go
raffaelespazzoli: because it needs to be a string...so quotes

api/v1alpha1/commons.go
raffaelespazzoli: from the docs:
```
A default value will be accepted as any value valid for the field. Formatting for common types include: boolean: true, string: Cluster, numerical: 1.24, array: {1,2}, object: {policy: &#34;delete&#34;}). Defaults should be defined in pruned form, and only best-effort validation will be performed. Full validation of a default requires submission of the containing CRD to an apiserver.
```

PR #6, opened 2021-11-02
Vault PKI Secret Engine

eye0fra:
- [X] PKI Engine implementation
    - [X] PKISecretEngineConfig Root CA
    - [X] /my-pki/root/generate/internal -> PKISecretEngineConfig CR with type root
    - [X] /my-pki/config/urls -> PKISecretEngineConfig CR

    - [X] PKISecretEngineConfig intermediate CA
    - [X]  /my-pki/intermediate/generate/internal -> PKISecretEngineConfig CR with type intermediate
    - [X]  /my-pki/config/urls -> PKISecretEngineConfig CR
    - [x]  /my-int/intermediate/set-signed PKISecretEngineConfig CR retrieved from a secret

- [X]  PKISecretEngineRole CR
   - [X]  /my-pki/roles/ 
- [X] PKI Engine sample
- [X] Test
- [X] Document
- [X] Improved local development.
- [x] PKI Engine Clean up logic


raffaelespazzoli:
for each secret engine there should be two CRDs: a xxxConfig and a xxxRole. Looking at the pki secret engine it seems that it could fit this mold even if the config endpoint is more complicated as it has a few sub endpoints. But because the relation between main config endpoint and sub endpoints is 1:1, a single xxxConfig CRD should suffice. The `intermediate` endpoint is something to be studied, I don't fully understand that use case and maybe we can leave it aside for now. The role endpoint is pretty standard and similar to the other secret engines.

eye0fra:
The `intermediate`is always a pki engine, check: [One CA one secret engine](https://www.vaultproject.io/docs/secrets/pki#one-ca-certificate-one-secrets-engine)

Here is how the intermediate chain works: https://learn.hashicorp.com/tutorials/vault/pki-engine

Ok, I will adjust with xxxConfig and xxxRole pattern.
Thanks for the review.

eye0fra:
Hey @raffaelespazzoli,

The PKI Engine API works slightly differently from the others, so it requires a different approach to the current generic implementation. Keeping it like that will become quite lengthy to cover the whole PKI API.
Nevertheless, I implemented the minimal endpoint to request a valid CA.
Let me know if you like to discuss how to implement multiple paths per API!

eye0fra:
```yml 
kind: PKIEngineMount
metadata:
  name: pki-sample
spec:
  authentication: 
    path: kubernetes
    role: pki-engine-admin
  path: test-vault-config-operator
  # Engine Tune
  tune:
    maxLeaseTTL: "8760h"
  # Creates PKISecretEngineRoot 
  # API /<path>/root/generate/:type
  root:
    commonName: pki-vault-demo.internal.io
    TTL: "8760h"
  # Creates PKISecretEngineIntermediate
  # API /<path>/intermediate/generate/:type
  # mutually exclusive with root, you can't create both.
  intermediate:
    commonName: intermediate.pki-vault-demo.internal.io
    TTL: "8760h"
    type: internal
    sign:
      # API /<path>/intermediate/set-signed
      internal:
        # API /<path-root>/root/sign-intermediate
        root: test-vault-config-operator
      external:
        secretRef:
           # Operator should wait until the external-signed secret contains the signed certificate
           name: external-signed
  # Creates PKISecretEngineRole
  # name is the metadata.name
  # API /<path>/roles/:name
  role:
    allowedDomains: 
    - internal.io
    - pki-vault-demo.svc
    - example.com
    allowSubdomains: "true"
    allowedOtherSans: "*"
    allowGlobDomains: "true"
    allowedUriSans: 
    - "*-pki-vault-demo.apps.example.com"
    maxTtl: "8760h" 
  # Creates PKISecretEngineConfig
  # name path is the metadata.name
  # API /<path>/config/urls
  config:
    issuingCertificates: 
    - https://vault-internal.vault.svc:8200/v1/test-vault-config-operator/ca
    crlDistributionPoints: 
    - https://vault-internal.vault.svc:8200/v1/test-vault-config-operator/crl"

```

raffaelespazzoli (commented):
eye0fra, thanks for the PR. Look very good already. I dropped some comments. If you disagree on the way to use defaults, let's talk.

api/v1alpha1/pkisecretengineconfig_types.go
```diff
@@ -0,0 +1,539 @@
...
+
+	// Specifies custom OID/UTF8-string SANs. These must match values specified on the role in allowed_other_sans (see role creation for allowed_other_sans globbing rules). The format is the same as OpenSSL: <oid>;<type>:<value> where the only current valid type is UTF8. This can be a comma-delimited list or a JSON string slice.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=""
+	OtherSans string `json:"otherSans,omitempty"`
+
+	// Specifies the requested Time To Live (after which the certificate will be expired). This cannot be larger than the engine's max (or, if not set, the system max).
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=""
+	TTL string `json:"TTL,omitempty"`
```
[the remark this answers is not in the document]
eye0fra: It should be Duration like the others thanks for spotting it.

api/v1alpha1/pkisecretengineconfig_types.go
```diff
@@ -0,0 +1,539 @@
...
+	PKICommon `json:",inline"`
+
+	PKIConfig `json:",inline"`
+
+	PKIIntermediate `json:",inline"`
+}
+
+type PKIType struct {
+	// Specifies the type of certificate authority. Root CA or Intermediate CA. This is part of the request URL.
+	// +kubebuilder:validation:Required
```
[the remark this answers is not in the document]
eye0fra: Good point!

api/v1alpha1/utils/vaultutils.go
```diff
@@ -40,7 +48,7 @@ func read(context context.Context, path string) (map[string]interface{}, bool, e
 	secret, err := vaultClient.Logical().Read(path)
 	if err != nil {
 		if respErr, ok := err.(*vault.ResponseError); ok {
-			if respErr.StatusCode == 404 {
+			if respErr.StatusCode == 404 || respErr.StatusCode == 204 {
```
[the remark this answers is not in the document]
eye0fra: 204 is No Content. I will use it to check if the CA is already generated `/pki/ca/pem` in case there is a misalignment between the vault and the CR, like manual deletion of the root CA.

controllers/pkisecretengineconfig_controller.go
```diff
@@ -0,0 +1,83 @@
...
+type PKISecretEngineConfigReconciler struct {
+	util.ReconcilerBase
+	Log            logr.Logger
+	ControllerName string
+}
+
+// //+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=redhatcop.redhat.io,resources=pkisecretengineconfigs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=redhatcop.redhat.io,resources=pkisecretengineconfigs/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=redhatcop.redhat.io,resources=pkisecretengineconfigs/finalizers,verbs=update
```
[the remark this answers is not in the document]
eye0fra: Good point.

api/v1alpha1/pkisecretengineconfig_types.go
```diff
@@ -0,0 +1,539 @@
...
+	PrivateKeyFormat string `json:"privateKeyFormat,omitempty"`
+
+	// Specifies the desired key type; must be rsa or ec.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="rsa"
+	KeyType string `json:"keyType,omitempty"`
+
+	// Specifies the number of bits to use. This must be changed to a valid value if the key_type is ec, e.g., 224, 256, 384 or 521.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=2048
```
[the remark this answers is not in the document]
eye0fra: well it depends of the key type the min and max, so probably we can add the validation on the webhook.

api/v1alpha1/pkisecretengineconfig_types.go
```diff
@@ -0,0 +1,539 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	ExcludeCnFromSans bool `json:"excludeCnFromSans,omitempty"`
+
+	// A comma separated string (or, string array) containing DNS domains for which certificates are allowed to be issued or signed by this CA certificate. Note that subdomains are allowed, as per RFC.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=""
+	PermittedDnsDomains string `json:"permittedDnsDomains,omitempty"`
+
+	// Specifies the OU (OrganizationalUnit) values in the subject field of issued certificates. This is a comma-separated string or JSON array.
```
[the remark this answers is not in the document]
eye0fra: Probably it is an array

api/v1alpha1/pkisecretengineconfig_types.go
```diff
@@ -0,0 +1,539 @@
...
+type PKIIntermediate struct {
+	// Secret retrieves the signed intermediate certificate from a Kubernetes secret. Allows submitting the signed CA certificate corresponding to a private key generated.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={}
+	Secret *corev1.LocalObjectReference `json:"secret,omitempty"`
+
+	// CertificateKey key to be used when retrieving the signed certificate
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="certificate"
+	CertificateKey string `json:"certificateKey,omitempty"`
```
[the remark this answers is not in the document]
eye0fra: it is the key name of the k8s secret.

When you say Vault secret you mean to add the functionality to retrieve from a Vault Secret instead of k8s?

api/v1alpha1/pkisecretengineconfig_types.go
```diff
@@ -0,0 +1,539 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	CrlDisable bool `json:"crlDisable,omitempty"`
+}
+
+type PKIIntermediate struct {
+	// Secret retrieves the signed intermediate certificate from a Kubernetes secret. Allows submitting the signed CA certificate corresponding to a private key generated.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={}
+	Secret *corev1.LocalObjectReference `json:"secret,omitempty"`
```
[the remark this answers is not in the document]
eye0fra: Yes, we could implement it.

eye0fra:
> eye0fra, thanks for the PR. Look very good already. I dropped some comments. If you disagree on the way to use defaults, let's talk.

Yes, you're right about the default, I noticed it. I will remove it so the CR will more readable.

api/v1alpha1/pkisecretengineconfig_types.go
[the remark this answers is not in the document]
raffaelespazzoli: #1 so for tls kube secrets the key of the certificate key is always `tls.key`.
#2 yes.

api/v1alpha1/pkisecretengineconfig_types.go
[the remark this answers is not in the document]
eye0fra: The default type is rsa and the default length for rsa is 2048.

Tiltfile
```diff
@@ -21,5 +21,8 @@ custom_build(
 )
 
 allow_k8s_contexts(k8s_context())
+watch_settings(ignore='config/*')
```
raffaelespazzoli, line 24: why this? I do use the fact that if these files change the kustomize has to be rerun...

api/v1alpha1/commons.go
```diff
@@ -159,6 +159,13 @@ func cleansePath(path string) string {
 	return strings.Trim(strings.ReplaceAll(path, "//", "/"), "/")
 }
 
+func toString(name interface{}) string {
```
raffaelespazzoli, line 162: I need to understand what this is about. face value it does not look good because an interface cannot be safely converted to a string and there is no error management.

api/v1alpha1/pkisecretengineconfig_types.go
```diff
@@ -0,0 +1,535 @@
...
+	SerialNumber string `json:"serialNumber,omitempty"`
+}
+
+type PKIConfig struct {
+	// +kubebuilder:validation:Optional
+	PKIConfigUrls `json:",inline"`
+	// +kubebuilder:validation:Optional
+	PKIConfigCRL `json:",inline"`
+	// // +kubebuilder:validation:Optional
+	// PKIConfigCA `json:",inline"`
```
raffaelespazzoli, line 173: if this is not needed, can we remove it?

api/v1alpha1/pkisecretengineconfig_types.go
```diff
@@ -0,0 +1,535 @@
...
+	// +kubebuilder:default="tls.crt"
+	CertificateKey string `json:"certificateKey,omitempty"`
+
+	// Use the configured refered Vault PKISecretEngineConfig to issue a certificate with appropriate values for acting as an intermediate CA.
+	// +kubebuilder:validation:Optional
+	InternalSign *corev1.LocalObjectReference `json:"internalSign,omitempty"`
+
+	CSR string `json:"-"`
+
+	SignedIntermediate string `json:"-"`
```
raffaelespazzoli, line 230: if this is a private field, it should not be exported.

api/v1alpha1/pkisecretengineconfig_types.go
```diff
@@ -0,0 +1,535 @@
...
+	// CertificateKey key to be used when retrieving the signed certificate
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="tls.crt"
+	CertificateKey string `json:"certificateKey,omitempty"`
+
+	// Use the configured refered Vault PKISecretEngineConfig to issue a certificate with appropriate values for acting as an intermediate CA.
+	// +kubebuilder:validation:Optional
+	InternalSign *corev1.LocalObjectReference `json:"internalSign,omitempty"`
+
+	CSR string `json:"-"`
```
raffaelespazzoli, line 228: if this is a private field, it should not be exported.

api/v1alpha1/pkisecretengineconfig_types.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+	return nil
+}
+
+func (p *PKISecretEngineConfig) GetExportedPayload(data map[string]interface{}) map[string]string {
+
+	payload := map[string]string{}
+
+	if p.Spec.Type == "root" {
+		payload["issuing_ca"] = toString(data["issuing_ca"])
```
raffaelespazzoli, line 339: ok I see where you are using to string. why did you have to do it this way?

api/v1alpha1/pkisecretengineconfig_types.go
```diff
@@ -0,0 +1,535 @@
...
+func (i *PKIConfigCRL) toMap() map[string]interface{} {
+	payload := map[string]interface{}{}
+
+	payload["expiry"] = i.CRLExpiry
+	payload["disable"] = i.CRLDisable
+
+	return payload
+}
+
+// func (i *PKIConfigCA) toMap() map[string]interface{} {
```
raffaelespazzoli, line 529: if this is not needed can we remove it?

api/v1alpha1/utils/vaultutils.go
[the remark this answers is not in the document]
raffaelespazzoli: mmm, this will change the behavior for everyone correct? I'm not sure it's right.

config/local-development/tilt/env-replace-image.yaml
```diff
@@ -1,4 +1,16 @@
-- op: replace
-  path: /spec/template/spec/containers/1/image
-  value:
-    quay.io/$repo/vault-config-operator:latest
+apiVersion: apps/v1
```
raffaelespazzoli, line 1: mm ok this is new and I am confused. maybe we can meet to discuss this

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
+  newName: quay.io/mmascia/vault-config-operator
```
raffaelespazzoli, line 15: this is not right and if you felt you needed to do it there is something wrong

config/manifests/bases/vault-config-operator.clusterserviceversion.yaml
```diff
@@ -90,6 +114,11 @@ spec:
       kind: SecretEngineMount
       name: secretenginemounts.redhatcop.redhat.io
       version: v1alpha1
+    - description: VaultSecret is the Schema for the vaultsecrets API
```
raffaelespazzoli, line 117: why is this being added? this should already have been here...

config/samples/redhatcop_v1alpha1_pkisecretengineconfig.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: redhatcop.redhat.io/v1alpha1
```
raffaelespazzoli, line 1: we should have real examples. I am not sure if I asked everyone else to do it. I will check, but please change this to a real example.

config/samples/redhatcop_v1alpha1_pkisecretenginerole.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: redhatcop.redhat.io/v1alpha1
```
raffaelespazzoli, line 1: we should have real examples. I am not sure if I asked everyone else to do it. I will check, but please change this to a real example.

test/pki-secret-engine-certificate-sample.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: cert-manager.io/v1
```
raffaelespazzoli, line 1: do you have documented the test sequence to run the manual test with cert-manager? I have not seen it.
There should be a readme in the test folder.

raffaelespazzoli (changes requested):
some changes and questions. The biggest is the one around tilt. Perhaps we should meet to discuss it.

Tiltfile
eye0fra: it is due to the control loop, that goes infinite :) every time a file change (make manifests), even if it is identical.
The first run works, but if you try from the UI to refresh the tiltfile you will see an infinite loop.
Try to comment out and see it.
Kustomize will always run so and will get any changes.

api/v1alpha1/commons.go
eye0fra: I need it here api/v1alpha1/pkisecretengineconfig_types.go on the GetExportedPayload.
The type of vault secret data is `map[string]interface{}`
For specific field is "guaranteed" that is a string type.
We could check the typeof before applying the cast. 
Do you have a better way of handling it?

api/v1alpha1/pkisecretengineconfig_types.go
eye0fra: what is the better approach? can't sign an interface to a string.

api/v1alpha1/pkisecretengineconfig_types.go
eye0fra: is it not the correct syntax? Or just remove the JSON part?

config/manager/kustomization.yaml
eye0fra: no forgot to roll back it.

config/local-development/tilt/env-replace-image.yaml
eye0fra: sure

config/samples/redhatcop_v1alpha1_pkisecretengineconfig.yaml
eye0fra: why is not real?

test/pki-secret-engine-certificate-sample.yaml
eye0fra: I will document it

config/samples/redhatcop_v1alpha1_pkisecretenginerole.yaml
eye0fra: why is not real?

config/manifests/bases/vault-config-operator.clusterserviceversion.yaml
eye0fra: Too many rebase :P, apart joke mmm maybe was never committed?

test/pki-secret-engine-certificate-sample.yaml
eye0fra: ok, shall I move it in a dedicated directory for PKI engine with the readme.md?

raffaelespazzoli:
I reviewed this and I think it's ready to be merged. just one change: can you add the pki engine config and role links in the readme together with all the other APIs.

PR #12, opened 2021-11-26
RabbitMQ Engine

rakivnenko1994:


api/v1alpha1/kubernetesauthenginerole_types.go
```diff
@@ -127,7 +127,6 @@ func (r *KubernetesAuthEngineRole) IsValid() (bool, error) {
 type VRole struct {
 
 	// TargetServiceAccounts is a list of service account names that will receive this role
-	// +kubebuilder:validation:Required
```
raffaelespazzoli, line 130: why did you remove this?

api/v1alpha1/rabbitmqsecretengineconfig_types.go
```diff
@@ -0,0 +1,237 @@
...
+	RMQSEConfig `json:",inline"`
+
+	// RootCredentials specifies how to retrieve the credentials for this RabbitMQEngine connection.
+	// +kubebuilder:validation:Required
+	RootCredentials RootCredentialConfig `json:"rootCredentials,omitempty"`
+}
+
+type RMQSEConfig struct {
+	// ConnectionURL Specifies the connection string used to connect to the RabbitMQ cluster.
+	// +kubebuilder:validation:Required
```
raffaelespazzoli, line 58: this is a good place to add a regexp validation if there is a good one for a rabbitmq connection string.

api/v1alpha1/rabbitmqsecretengineconfig_types.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func init() {
+	SchemeBuilder.Register(&RabbitMQSecretEngineConfig{}, &RabbitMQSecretEngineConfigList{})
+}
+
+func (fields *RMQSEConfig) rabbitMQToMap() map[string]interface{} {
+	payload := map[string]interface{}{}
+	payload["connection_uri"] = fields.ConnectionURI
+	payload["verify_connection"] = &fields.VerifyConnection
```
raffaelespazzoli, line 127: is the reference `&` needed here?

api/v1alpha1/rabbitmqsecretengineconfig_webhook.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (r *RabbitMQSecretEngineConfig) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+
+//+kubebuilder:webhook:path=/mutate-redhatcop-redhat-io-v1alpha1-rabbitmqsecretengineconfig,mutating=true,failurePolicy=fail,sideEffects=None,groups=redhatcop.redhat.io,resources=rabbitmqsecretengineconfigs,verbs=create;update,versions=v1alpha1,name=mrabbitmqsecretengineconfig.kb.io,admissionReviewVersions={v1,v1beta1}
```
raffaelespazzoli, line 39: if there is not default to set, then we should not have this webhook. Also usually this is needed only at create time

api/v1alpha1/rabbitmqsecretengineconfig_webhook.go
```diff
@@ -0,0 +1,79 @@
...
+var _ webhook.Defaulter = &RabbitMQSecretEngineConfig{}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (r *RabbitMQSecretEngineConfig) Default() {
+	rabbitmqsecretengineconfiglog.Info("default", "name", r.Name)
+	// Doesn't have any resources to be deleted.
+}
+
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
+//+kubebuilder:webhook:path=/validate-redhatcop-redhat-io-v1alpha1-rabbitmqsecretengineconfig,mutating=false,failurePolicy=fail,sideEffects=None,groups=redhatcop.redhat.io,resources=rabbitmqsecretengineconfigs,verbs=create;update,versions=v1alpha1,name=vrabbitmqsecretengineconfig.kb.io,admissionReviewVersions={v1,v1beta1}
```
raffaelespazzoli, line 50: probably this is needed only at update time.

api/v1alpha1/rabbitmqsecretenginerole_types.go
```diff
@@ -0,0 +1,176 @@
...
+type RMQSERole struct {
+	// Comma-separated RabbitMQ permissions tags to associate with the user. This determines the level of
+	// access to the RabbitMQ management UI granted to the user. Omitting this field will
+	// lead to a user than can still connect to the cluster through messaging protocols,
+	// but cannot perform any management actions.
+	// +kubebuilder:validation:Optional
+	Tags string `json:"tags,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	Vhosts []Vhost `json:"vhosts,omitempty"`
```
raffaelespazzoli, line 55: when array are used see if any of the processing directive is applicable: https://book.kubebuilder.io/reference/markers/crd-processing.html
this comment referes also to the next field.

api/v1alpha1/rabbitmqsecretenginerole_types.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (rabbitMQ *RabbitMQSecretEngineRole) PrepareInternalValues(context context.Context, object client.Object) error {
+	return nil
+}
+
+func (rabbitMQ *RabbitMQSecretEngineRole) IsValid() (bool, error) {
+	return true, nil
+}
+
+func convertToJson(vhosts interface{}) (string) {
```
raffaelespazzoli, line 162: for this to work the json names used in the CRD must correspond to what Vault expects. This was usually not the case for me as CRD use camelcase and vault uses underscores... I had to define a xxxToMap for each nested struct.

api/v1alpha1/rabbitmqsecretenginerole_webhook.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (r *RabbitMQSecretEngineRole) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+
+//+kubebuilder:webhook:path=/mutate-redhatcop-redhat-io-v1alpha1-rabbitmqsecretenginerole,mutating=true,failurePolicy=fail,sideEffects=None,groups=redhatcop.redhat.io,resources=rabbitmqsecretengineroles,verbs=create;update,versions=v1alpha1,name=mrabbitmqsecretenginerole.kb.io,admissionReviewVersions={v1,v1beta1}
```
raffaelespazzoli, line 40: should be done only at create time

api/v1alpha1/rabbitmqsecretenginerole_webhook.go
```diff
@@ -0,0 +1,82 @@
...
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (r *RabbitMQSecretEngineRole) Default() {
+	rabbitmqsecretenginerolelog.Info("default", "name", r.Name)
+	if !controllerutil.ContainsFinalizer(r, GetFinalizer(r)) {
+		controllerutil.AddFinalizer(r, GetFinalizer(r))
+	}
+}
+
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
+//+kubebuilder:webhook:path=/validate-redhatcop-redhat-io-v1alpha1-rabbitmqsecretenginerole,mutating=false,failurePolicy=fail,sideEffects=None,groups=redhatcop.redhat.io,resources=rabbitmqsecretengineroles,verbs=create;update,versions=v1alpha1,name=vrabbitmqsecretenginerole.kb.io,admissionReviewVersions={v1,v1beta1}
```
raffaelespazzoli, line 53: this is probably needed only at update time.

controllers/rabbitmqsecretengineconfig_controller.go
```diff
@@ -0,0 +1,130 @@
...
+
+// RabbitMQSecretEngineConfigReconciler reconciles a RabbitMQSecretEngineConfig object
+type RabbitMQSecretEngineConfigReconciler struct {
+	util.ReconcilerBase
+	Log            logr.Logger
+	ControllerName string
+}
+
+//+kubebuilder:rbac:groups=redhatcop.redhat.io,resources=rabbitmqsecretengineconfigs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=redhatcop.redhat.io,resources=rabbitmqsecretengineconfigs/status,verbs=get;update;patch
```
raffaelespazzoli, line 44: events seems to be missing here, possibly something else like secrets....

controllers/rabbitmqsecretengineconfig_controller.go
```diff
@@ -0,0 +1,130 @@
...
+func (r *RabbitMQSecretEngineConfigReconciler) manageReconcileLogic(context context.Context, instance client.Object) error {
+	log := log.FromContext(context)
+	vaultEndpoint := vaultutils.NewVaultEndpoint(instance)
+	// prepare internal values
+	err := instance.(vaultutils.VaultObject).PrepareInternalValues(context, instance)
+	if err != nil {
+		log.Error(err, "unable to prepare internal values", "instance", instance)
+		return err
+	}
+	err = vaultEndpoint.Create(context)
```
raffaelespazzoli, line 102: should this be CreateorUpdate?

controllers/rabbitmqsecretengineconfig_controller.go
```diff
@@ -0,0 +1,130 @@
...
+// SetupWithManager sets up the controller with the Manager.
+func (r *RabbitMQSecretEngineConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	filter := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return true
+		},
+		CreateFunc: func(e event.CreateEvent) bool {
+			return true
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
```
raffaelespazzoli, line 119: why are you filtering out the delete event? you need it to run the finalizer...

raffaelespazzoli (commented):
Vladislav thanks for your PR. This is a very good start. I left some comments.

api/v1alpha1/rabbitmqsecretengineconfig_types.go
rakivnenko1994: Sure, added simple regex  ^(http|https):\/\/.+$ for this, don't think anything more complex needed here

api/v1alpha1/kubernetesauthenginerole_types.go
rakivnenko1994: To be able to use default value, removed default instead

api/v1alpha1/rabbitmqsecretengineconfig_webhook.go
rakivnenko1994: Removed.

api/v1alpha1/rabbitmqsecretengineconfig_webhook.go
rakivnenko1994: Why do you think it's not useful for create events?

api/v1alpha1/rabbitmqsecretenginerole_types.go
rakivnenko1994: Added for arrays:
	// +listType=map
	// +listMapKey=vhostName
	// +listMapKey=permissions
	// +kubebuilder:validation:UniqueItems=true

controllers/rabbitmqsecretengineconfig_controller.go
rakivnenko1994: Secrets are listed by other controllers, do you still want me to add them for readability reasons?

controllers/rabbitmqsecretengineconfig_controller.go
rakivnenko1994: Update part requires to be able to read configuration, RabbitMQ engine connection configuration is not readable, as it contain secrets as a plain text

controllers/rabbitmqsecretengineconfig_controller.go
rakivnenko1994: There is no ability to read or delete this engine configuration. Should I remove whole finalizer part for it?

api/v1alpha1/rabbitmqsecretenginerole_webhook.go
rakivnenko1994: Updated

api/v1alpha1/rabbitmqsecretengineconfig_webhook.go
rakivnenko1994: Updated

api/v1alpha1/rabbitmqsecretenginerole_webhook.go
rakivnenko1994: Updated

api/v1alpha1/rabbitmqsecretengineconfig_types.go
rakivnenko1994: Removed.

api/v1alpha1/rabbitmqsecretenginerole_types.go
rakivnenko1994: Faced an issue with list type map, it requires default value and as this arrays can be empty it's not acceptable. UniqueItems cannot be set to true since the runtime complexity becomes quadratic.
Added // +listType=atomic instead.

api/v1alpha1/rabbitmqsecretenginerole_types.go
rakivnenko1994: Updated json convertion

raffaelespazzoli:
I thought about this a little and I'd say for now let's do the following:
1. deleting a config is a no-op on Vault, it of course deletes the object in Kube
2. in the validation webhook, let's ensure that someone cannot define two config for the same reabbit mq engine endpoint endpoint

api/v1alpha1/rabbitmqsecretengineconfig_types.go
```diff
@@ -0,0 +1,260 @@
...
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// RabbitMQSecretEngineConfigSpec defines the desired state of RabbitMQSecretEngineConfig
+type RabbitMQSecretEngineConfigSpec struct {
+	// Authentication is the k8s auth configuration to be used to execute this request
+	// +kubebuilder:validation:Required
+	Authentication KubeAuthConfiguration `json:"authentication,omitempty"`
+
+	// Path at which to make the configuration.
+	// The final path will be {[spec.authentication.namespace]}/{spec.path}/config/{metadata.name}.
```
raffaelespazzoli, line 43: the doc here is wrong, I feel like it should say that the final path is 
`{[spec.authentication.namespace]}/{spec.path}/{metadata.name}/config/connection`
but you tell me.
The `getPath`, should behave consistently with the doc.

api/v1alpha1/rabbitmqsecretengineconfig_types.go
raffaelespazzoli: this is good. I am concerned that in the case of a https connection one cannot pass the CA. it seems like a gap in the feature of this secret engine.

api/v1alpha1/kubernetesauthenginerole_types.go
rakivnenko1994: When I'm nit providing targetServiceAccounts and both required true with default={"default"}, I'm getting:
`error: error validating "./test/kube-auth-engine-role.yaml": error validating data: ValidationError(KubernetesAuthEngineRole.spec): missing required field "targetServiceAccounts" in io.redhat.redhatcop.v1alpha1.KubernetesAuthEngineRole.spec; if you choose to ignore these errors, turn validation off with --validate=false`
`error: error validating "./test/rabbitmq-engine-admin-role.yaml": error validating data: ValidationError(KubernetesAuthEngineRole.spec): missing required field "targetServiceAccounts" in io.redhat.redhatcop.v1alpha1.KubernetesAuthEngineRole.spec; if you choose to ignore these errors, turn validation off with --validate=false`
So, I'm removing this field as default to prevent further confusion.
Changed line number to clearly show that default is removed.

api/v1alpha1/rabbitmqsecretengineconfig_types.go
rakivnenko1994: Not sure which CA you would like to pass. If you are talking about connection between Vault Engine and RabbitMQ cluster there is no way to pass CA cert, it's expected for RabbitMQ to be configured with proper CA cert.

api/v1alpha1/rabbitmqsecretengineconfig_types.go
rakivnenko1994: Updated

PROJECT
```diff
@@ -148,4 +148,30 @@ resources:
     defaulting: true
     validation: true
     webhookVersion: v1
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: redhat.io
+  group: redhatcop
+  kind: RabbitMQSecretEngineRole
```
raffaelespazzoli, line 157: soemthing does not look right here as RabbitMQSecretEngineRole appears twice

api/v1alpha1/rabbitmqsecretengineconfig_types.go
raffaelespazzoli: ok

api/v1alpha1/rabbitmqsecretengineconfig_webhook.go
```diff
@@ -0,0 +1,100 @@
...
+		rabbitmqsecretengineconfiglog.Error(err, "unable to create vault client required for instance validation", "instance", r)
+		return err
+	}
+	mounts, err := vaultClient.Sys().ListMounts()
+	if err != nil {
+		// Operator requires permission to list mounts for validation
+		rabbitmqsecretengineconfiglog.Error(err, "failed to list mounts", "instance", r)
+		return err
+	}
+	if mountExists := checkIfMountExists(mounts, string(r.Spec.Path)); mountExists {
```
raffaelespazzoli, line 61: I am confused by this validation logic. We want the mount to exists..but the logic seems to be erroring out if it exists.
Perhaps you wanted to check if it was already configured?

controllers/rabbitmqsecretengineconfig_controller.go
raffaelespazzoli: yes please

docs/api.md
```diff
@@ -13,8 +13,13 @@ This section of the documentation provides high-level documentation on the suppo
   - [DatabaseSecretEngineConfig](#databasesecretengineconfig)
   - [DatabaseSecretEngineRole](#databasesecretenginerole)
   - [RandomSecret](#randomsecret)
+<<<<<<< HEAD
```
raffaelespazzoli, line 16: this is a left over from a merge

raffaelespazzoli (changes requested):
added some more comments. I think we are close, most of the comment s are minor. But I'd like to learn more about the need for a lease.

docs/api.md
rakivnenko1994: Removed.

PROJECT
rakivnenko1994: Fixed.

controllers/rabbitmqsecretengineconfig_controller.go
rakivnenko1994: Done

api/v1alpha1/utils/vaultobject.go
```diff
@@ -80,3 +80,36 @@ func (ve *VaultEndpoint) CreateOrUpdate(context context.Context) error {
 	}
 	return nil
 }
+
+type VaultObjectWithLease interface {
```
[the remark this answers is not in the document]
rakivnenko1994: This endpoint configures the lease settings for generated credentials from https://www.vaultproject.io/api-docs/secret/rabbitmq#configure-lease.
I tried to use tune, but it seems to have no effect. I set tune default_lease_ttl to 80000s and lease TTL is 86400, but when I generate credentials there "lease_duration" is 86400s.
I'm creating a separate object to be able to set separate path and values, so it will be updated as part of a single reconcile and TTL values can be provided as part of engine configuration.

api/v1alpha1/kubernetesauthenginerole_types.go
raffaelespazzoli: let's reinstate the default

api/v1alpha1/rabbitmqsecretengineconfig_webhook.go
raffaelespazzoli: we can skip this validation, up to you.

raffaelespazzoli: can we rename this RabbitMQEngineConfigVaultObject?

rakivnenko1994: Renamed and updated usage in controller

rakivnenko1994:
Resolved comments. Updated controller to utilize single Vault object and added creation validation for RabbitMQ configs.

api/v1alpha1/commons.go
```diff
@@ -175,6 +175,68 @@ type VaultSecretReference struct {
...
+		return errors.New("only one of spec.rootCredentials.vaultSecret or spec.rootCredentials.secret or spec.rootCredentials.randomSecret can be specified")
+	}
+	return nil
+}
+
 func GetFinalizer(instance client.Object) string {
 	return "controller-" + strings.ToLower(instance.GetObjectKind().GroupVersionKind().Kind)
 }
+
+func GetVaultKV2Secret(path string, context context.Context) (*vault.Secret, error) {
```
raffaelespazzoli, line 229: this should be called `GetVaultSecret`. Also I'm pretty sure we already have this function....vaultiutils.readSecret or vaultutils.Read.
While you are at it can you refacrtor read to call readSecret? I should have caught it in the last merge.

api/v1alpha1/databasesecretengineconfig_types.go
```diff
@@ -124,9 +92,8 @@ func (r *DatabaseSecretEngineConfig) setInternalCredentials(context context.Cont
 			log.Error(err, "unable to retrieve RandomSecret", "instance", r)
 			return err
 		}
-		secret, err := vaultClient.Logical().Read(randomSecret.GetPath())
+		secret, err := GetVaultKV2Secret(randomSecret.GetPath(), context)
 		if err != nil {
-			log.Error(err, "unable to retrieve vault secret", "instance", r)
```
raffaelespazzoli, line 129: I like to print error every time event if there is nothing to add...

api/v1alpha1/databasesecretengineconfig_types.go
```diff
@@ -150,9 +117,8 @@ func (r *DatabaseSecretEngineConfig) setInternalCredentials(context context.Cont
 		return nil
 	}
 	if r.Spec.RootCredentials.VaultSecret != nil {
-		secret, err := vaultClient.Logical().Read(string(r.Spec.RootCredentials.VaultSecret.Path))
+		secret, err := GetVaultKV2Secret(string(r.Spec.RootCredentials.VaultSecret.Path), context)
 		if err != nil {
-			log.Error(err, "unable to retrieve vault secret", "instance", r)
```
raffaelespazzoli, line 155: I like to print error every time event if there is nothing to add...

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
```diff
@@ -151,7 +244,7 @@ metadata:
       Vault.
     operatorframework.io/suggested-namespace: vault-config-operator
     operators.openshift.io/infrastructure-features: '["Disconnected"]'
-    operators.operatorframework.io/builder: operator-sdk-v1.13.0+git
+    operators.operatorframework.io/builder: operator-sdk-v1.13.1
```
raffaelespazzoli, line 247: can you add this: `operatorframework.io/cluster-monitoring: "true"` here? It should have been there.

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
```diff
@@ -235,6 +345,16 @@ spec:
     spec:
       clusterPermissions:
       - rules:
+        - apiGroups:
```
raffaelespazzoli, line 348: not sure what is going on here, but these will be added during the bundle build and should not be here

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
```diff
@@ -393,6 +513,32 @@ spec:
           - get
           - patch
           - update
+        - apiGroups:
```
raffaelespazzoli, line 516: not sure what is going on here, but these will be added during the bundle build and should not be here

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
```diff
@@ -419,6 +565,52 @@ spec:
...
```
raffaelespazzoli, line 568: not sure what is going on here, but these will be added during the bundle build and should not be here

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
```diff
@@ -505,7 +697,7 @@ spec:
                 - --v=10
                 - --tls-cert-file=/etc/certs/tls/tls.crt
                 - --tls-private-key-file=/etc/certs/tls/tls.key
-                image: registry.redhat.io/openshift4/ose-kube-rbac-proxy@sha256:6d57bfd91fac9b68eb72d27226bc297472ceb136c996628b845ecc54a48b31cb
+                image: quay.io/redhat-cop/kube-rbac-proxy:v0.11.0
```
raffaelespazzoli, line 700: this must have come from a rushed merge, all of this should not be here....

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
```diff
@@ -802,6 +1054,46 @@ spec:
     targetPort: 9443
     type: ValidatingAdmissionWebhook
     webhookPath: /validate-redhatcop-redhat-io-v1alpha1-randomsecret
+  - admissionReviewVersions:
```
raffaelespazzoli, line 1057: not sure what is going on here, but these will be added during the bundle build and should not be here

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
```diff
@@ -661,6 +853,66 @@ spec:
     targetPort: 9443
     type: MutatingAdmissionWebhook
     webhookPath: /mutate-redhatcop-redhat-io-v1alpha1-randomsecret
+  - admissionReviewVersions:
```
raffaelespazzoli, line 856: not sure what is going on here, but these will be added during the bundle build and should not be here

bundle.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
```
raffaelespazzoli, line 6: now I see the problem, please delete all of these files and ignore the bundle dir. This should not have been committed.

config/manifests/bases/vault-config-operator.clusterserviceversion.yaml
```diff
@@ -56,11 +56,28 @@ spec:
       kind: KubernetesAuthEngineRole
```
raffaelespazzoli, line 56: can you add this: operatorframework.io/cluster-monitoring: "true" here? It should have been there. In the annotations section

raffaelespazzoli (changes requested):
main thing is that the bundle dir is being committed. It should be ignored. then there are a few minor comments.

api/v1alpha1/databasesecretengineconfig_types.go
rakivnenko1994: Error is printed by ReadSecret method, it will a duplication of the error if I add it back

api/v1alpha1/databasesecretengineconfig_types.go
rakivnenko1994: Error is printed by ReadSecret method, it will a duplication of the error if I add it back

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
rakivnenko1994: Removed.

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
rakivnenko1994: Removed.

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
rakivnenko1994: Removed.

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
rakivnenko1994: Removed.

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
rakivnenko1994: Removed.

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
rakivnenko1994: Removed.

bundle/manifests/vault-config-operator.clusterserviceversion.yaml
rakivnenko1994: Removed.

bundle.Dockerfile
rakivnenko1994: Removed.

config/manifests/bases/vault-config-operator.clusterserviceversion.yaml
rakivnenko1994: Removed.

api/v1alpha1/commons.go
rakivnenko1994: Made readSecret a public method, replaced GetVaultSecret with it and updated ReadSecret to use new method

PR #13, opened 2021-11-29
Vaultsecret

trevorbox:


api/v1alpha1/vaultsecret_types.go
```diff
@@ -0,0 +1,173 @@
...
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// VaultSecretSpec defines the desired state of VaultSecret
+type VaultSecretSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Url of the Vault instance.
+	// +kubebuilder:validation:Required
+	Url string `json:"url,omitempty"`
```
raffaelespazzoli, line 37: this is not needed in this operator

api/v1alpha1/vaultsecret_types.go
```diff
@@ -0,0 +1,173 @@
...
+	// RefreshPeriod if specified, the operator will refresh the secret with the given frequency.
+	// Defaults to five minutes, and must be at least one minute.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="5m"
+	RefreshPeriod *metav1.Duration `json:"refreshPeriod,omitempty"`
```
raffaelespazzoli, line 42: I'd like to discuss this feature:
1. most vault secrets tells you when they are going to expire, so the operator should made the choice based on that info.
2.  if we want to support an interval-based refresh, then there should be the option of never update. And perhaps that should be the default.
3. shouldn't this be at the secret level?

api/v1alpha1/vaultsecret_types.go
```diff
@@ -0,0 +1,173 @@
...
+	// KVSecrets are the Key/Value secrets in Vault.
+	// +kubebuilder:validation:Required
+	KVSecrets []KVSecret `json:"kvSecrets,omitempty"`
```
raffaelespazzoli, line 45: why are these called KV secrets? They are references to Vault secrets correct?  look at the common types, there is already one that represents a reference to a Vault secret.

api/v1alpha1/vaultsecret_types.go
```diff
@@ -0,0 +1,173 @@
...
+	// TemplatizedK8sSecret is the formatted K8s Secret created by templating from the Vault KV secrets.
+	// +kubebuilder:validation:Required
+	TemplatizedK8sSecret TemplatizedK8sSecret `json:"output,omitempty"`
```
raffaelespazzoli, line 48: the json value seems wrong.

api/v1alpha1/vaultsecret_types.go
```diff
@@ -0,0 +1,173 @@
...
+	SchemeBuilder.Register(&VaultSecret{}, &VaultSecretList{})
+}
+
+type KVSecret struct {
+	// Name is an arbitrary, but unique, name for this KV Vault secret and referenced when templating.
+	// +kubebuilder:validation:Required
+	Name string `json:"name,omitempty"`
+	// Authentication is the kube aoth configuraiton to be used to execute this request
+	// +kubebuilder:validation:Required
+	Authentication KubeAuthConfiguration `json:"authentication,omitempty"`
```
raffaelespazzoli, line 102: supporting an different authentication per secret adds some complexity, are you sure it's needed? All other APIs of this operator have one top-level authentication field

api/v1alpha1/vaultsecret_types.go
```diff
@@ -0,0 +1,173 @@
...
+	// Keys is a list of keys to use for templating. If none are listed all keys are referenceable for templating.
+	// +kubebuilder:validation:Optional
+	Keys []string `json:"keys,omitempty"`
```
raffaelespazzoli, line 105: why this is needed? Why can't you just pass the entire secret as returned as a template parameter?

api/v1alpha1/vaultsecret_types.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+type TemplatizedK8sSecret struct {
+	// Name is the K8s Secret name to output to.
+	// +kubebuilder:validation:Required
+	Name string `json:"name,omitempty"`
+	// Type is the K8s Secret type to output to.
+	// +kubebuilder:validation:Required
+	Type string `json:"type,omitempty"`
+	// StringData is the K8s Secret stringData and allows specifying non-binary secret data in string form with go templating support
```
raffaelespazzoli, line 115: so the here the value is interpreted as a golang template, correct? Can you specify that in the doc.
There is decent support for golang templates in the utils library, hopefully you found it.

api/v1alpha1/vaultsecret_types.go
```diff
@@ -0,0 +1,173 @@
...
+	// to transform the Vault KV secrets into a formatted K8s Secret.
+	// The Sprig template library and Helm functions (like toYaml) are supported.
+	// +kubebuilder:validation:Required
+	StringData map[string]string `json:"stringData,omitempty"`
+	// Annotations are annotations to add to the final K8s Secret.
+	// +kubebuilder:validation:Optional
+	Annotations map[string]string `json:"annotations,omitempty"`
```
raffaelespazzoli, line 122: If annotaitons, the we also need labels...

api/v1alpha1/vaultsecret_webhook.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (r *VaultSecret) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+
+//+kubebuilder:webhook:path=/mutate-redhatcop-redhat-io-v1alpha1-vaultsecret,mutating=true,failurePolicy=fail,sideEffects=None,groups=redhatcop.redhat.io,resources=vaultsecrets,verbs=create;update,versions=v1alpha1,name=mvaultsecret.kb.io,admissionReviewVersions={v1,v1beta1}
```
raffaelespazzoli, line 38: this is probably needed only on create

config/local-development/tilt/replace-image.yaml
```diff
@@ -1,4 +1,4 @@
 - op: replace
-  path: /spec/template/spec/containers/1/image
+  path: /spec/template/spec/containers/0/image
```
raffaelespazzoli, line 2: please look at the pki pull request on they improved on the tilt solution, this should not be needed.

controllers/vaultsecret_controller.go
```diff
@@ -0,0 +1,226 @@
...
+// VaultSecretReconciler reconciles a VaultSecret object
+type VaultSecretReconciler struct {
+	util.ReconcilerBase
+	Log            logr.Logger
+	ControllerName string
+}
+
+//+kubebuilder:rbac:groups=redhatcop.redhat.io,resources=vaultsecrets,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=redhatcop.redhat.io,resources=vaultsecrets/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=redhatcop.redhat.io,resources=vaultsecrets/finalizers,verbs=update
```
raffaelespazzoli, line 51: events are probably needed here.

controllers/vaultsecret_controller.go
```diff
@@ -0,0 +1,226 @@
...
+
+	return nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *VaultSecretReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	needsCreation := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			newSecret, ok := e.ObjectNew.DeepCopyObject().(*redhatcopv1alpha1.VaultSecret)
```
raffaelespazzoli, line 200: I don't get this logic here....

pkg/template/funcs.go
```diff
@@ -0,0 +1,54 @@
+package template
```
raffaelespazzoli, line 1: we have all of these functions in the utils library...

pkg/template/template.go
```diff
@@ -0,0 +1,173 @@
+package template
```
raffaelespazzoli, line 1: also you should be able to reuse util for this logic

api/v1alpha1/vaultsecret_types.go
trevorbox: removed

api/v1alpha1/vaultsecret_types.go
trevorbox: The intention is to control the pull policy. It seems irrelevant to worry about the Vault configuration's expiration since users can update the secret at any time. I think keeping this pull interval completely separate is simple and makes the most sense to the user.

api/v1alpha1/vaultsecret_types.go
trevorbox: This is a vault Key Value Secret. No other data structure has what is needed from what I can tell. See the api documentation.

api/v1alpha1/vaultsecret_types.go
trevorbox: works fine.

config/local-development/tilt/replace-image.yaml
trevorbox: I did take a look and didn't see how this was resolved...

controllers/vaultsecret_controller.go
trevorbox: yes, added

controllers/vaultsecret_controller.go
trevorbox: Its fairly standard. checks if the spec was infact changed of the CR

pkg/template/funcs.go
trevorbox: ok, removed it.

pkg/template/template.go
trevorbox: removed.

api/v1alpha1/vaultsecret_types.go
```diff
@@ -97,12 +97,16 @@ type KVSecret struct {
 	// Name is an arbitrary, but unique, name for this KV Vault secret and referenced when templating.
 	// +kubebuilder:validation:Required
 	Name string `json:"name,omitempty"`
-	// Authentication is the kube aoth configuraiton to be used to execute this request
+	// Authentication is the kube auth configuraiton to be used to execute this request
```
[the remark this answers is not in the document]
trevorbox: done.

api/v1alpha1/vaultsecret_types.go
trevorbox: I sort of like the flexibility it allows.

api/v1alpha1/vaultsecret_types.go
trevorbox: ah ok - I was hoping the api supported fields like the cli. better security to only get the keys you need.

"However, according to [1], “field” is a cli feature. The api always returns a full response."
I'll remove this.

api/v1alpha1/vaultsecret_types.go
trevorbox: docs updated

api/v1alpha1/vaultsecret_types.go
trevorbox: labels added

api/v1alpha1/vaultsecret_webhook.go
trevorbox: why not update as well?

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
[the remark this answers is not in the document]
trevorbox: removed. allows debugging in vscode

api/v1alpha1/vaultsecret_types.go
raffaelespazzoli: you are thinking mostly about the kv secret store, but that is really a minor use case (although it's used in all of the demos because it's easy to set up), but all of the other secret engines have expiration. I'd like to discuss this more.

api/v1alpha1/vaultsecret_types.go
```diff
@@ -0,0 +1,170 @@
...
+	// Name is an arbitrary, but unique, name for this KV Vault secret and referenced when templating.
+	// +kubebuilder:validation:Required
+	Name string `json:"name,omitempty"`
+	// Authentication is the kube auth configuraiton to be used to execute this request
+	// +kubebuilder:validation:Required
+	Authentication KubeAuthConfiguration `json:"authentication,omitempty"`
+	// Path is the path of the secret.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default=kubernetes
+	Path Path `json:"path,omitempty"`
```
raffaelespazzoli, line 106: please add a comment that explains exactly what path will be used in Vault and what permission will be required (I assume read). Just like we do for the other APIs.

api/v1alpha1/vaultsecret_types.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+type TemplatizedK8sSecret struct {
+	// Name is the K8s Secret name to output to.
+	// +kubebuilder:validation:Required
+	Name string `json:"name,omitempty"`
+	// Type is the K8s Secret type to output to.
+	// +kubebuilder:validation:Required
+	Type string `json:"type,omitempty"`
```
raffaelespazzoli, line 115: this should reflect what is done in the secret API, do they have an enum or some validation on the possible values?

api/v1alpha1/vaultsecret_webhook.go
raffaelespazzoli: given that you don't do anything, it's actually not needed at all. What I meant with my comment is that usually here we set the finalized, which you do only at create time.

config/local-development/tilt/replace-image.yaml
raffaelespazzoli: I cannot accept this in this state, do you want to meet to fix it together?

controllers/vaultsecret_controller.go
```diff
@@ -0,0 +1,224 @@
...
+		mergedMap[kvSecret.Name] = data
+	}
+
+	k8sSecret, err := r.formatK8sSecret(instance, mergedMap)
+	if err != nil {
+		r.Log.Error(err, "unable to format k8s secret", "instance", instance)
+		return err
+	}
+
+	existingK8sSecret := &corev1.Secret{}
```
raffaelespazzoli, line 165: please use the createOrUpdate function from utils. This will also take care of controller reference

controllers/vaultsecret_controller.go
```diff
@@ -0,0 +1,224 @@
...
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return true
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
```
raffaelespazzoli, line 220: this is wrong as you are not watching the created secret.

controllers/vaultsecret_controller.go
```diff
@@ -0,0 +1,224 @@
...
+	now := metav1.NewTime(time.Now())
+	instance.Status.LastVaultSecretUpdate = &now
+
+	return nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *VaultSecretReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	needsCreation := predicate.Funcs{
```
raffaelespazzoli, line 196: there is n need to do this, the update event will be sent to you if there is a change anywhere, which is what you are checking here.

go.mod
```diff
@@ -7,12 +7,13 @@ require (
...
+	github.com/onsi/ginkgo v1.16.5
+	github.com/onsi/gomega v1.17.0
+	github.com/pkg/errors v0.9.1
+	github.com/redhat-cop/operator-utils v1.3.1
 	github.com/scylladb/go-set v1.0.2
-	k8s.io/api v0.21.2
-	k8s.io/apimachinery v0.21.2
-	k8s.io/client-go v0.21.2
-	sigs.k8s.io/controller-runtime v0.9.2
+	k8s.io/api v0.22.1
```
raffaelespazzoli, line 15: a downgrade makes me think that there's something misconfigured.

raffaelespazzoli (changes requested):

PR #15, opened 2022-01-14
VaultSecret

trevorbox:
This properly handles resyncing for dynamic secrets using a lease duration.


config/local-development/tilt/add-vault-environment-variables.yaml
```diff
@@ -1,5 +1,5 @@
 - op: add
-  path: /spec/template/spec/containers/1/env
```
raffaelespazzoli, line 2: does this still not work? I cannot accept this. After the new rebase this should have been fixed for everyone.

config/local-development/tilt/replace-image.yaml
```diff
@@ -1,4 +1,4 @@
 - op: replace
-  path: /spec/template/spec/containers/1/image
+  path: /spec/template/spec/containers/0/image
   value:
-    quay.io/raffaelespazzoli/vault-config-operator:latest
```
raffaelespazzoli, line 4: does this still not work? I cannot accept this. After the new rebase this should have been fixed for everyone.

controllers/vaultsecret_controller.go
```diff
@@ -0,0 +1,307 @@
...
+	instance.Status.LastVaultSecretUpdate = &now
+	instance.Status.VaultSecretDefinitionsStatus = definitionsStatus
+
+	return nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *VaultSecretReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	vaultSecretPredicate := predicate.Funcs{
```
raffaelespazzoli, line 258: this is not needed as what you have coded is the default behavior

controllers/vaultsecret_controller.go
```diff
@@ -0,0 +1,307 @@
...
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			r.Log.V(1).Info("Delete Event", "object", e.Object)
+			return true
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
+
+	k8sSecretPredicate := predicate.Funcs{
```
raffaelespazzoli, line 288: I don't understand this, why are you adding the `Owns` clause and then discarding all of the events?

docs/api.md
```diff
@@ -391,4 +392,90 @@ to read a new credential from this role, execute the following:
...
+
+> Note: if reading a dynamic secret you typically care to set the `refreshThreshold` only (not the `refreshPeriod`). For just Key/Value Vault secrets, set the `refreshPeriod`.
+> 
+> See https://www.vaultproject.io/docs/concepts/lease to understand lease durations.
+
+- `refreshPeriod` the pull interval for syncing Vault secrets with the K8s Secret. This settings takes precedence over any lease duration returned by vault, effectively controlling when exactly all vault secrets defined in the vaultSecretDefinitions should re-sync.
+- `refreshThreshold` this is will instruct the operator to refresh the K8s Secret when a percentage of the lease duration has elapsed, if no `refreshPeriod` is specified. This is particularly useful for controlling when dynamic secrets should be refreshed before the lease duration is exceeded. The default is 90, meaning the secret would refresh after 90% of the time has passed from the vault secret's lease duration. When multiple vaultSecretDefinitions are defined, the smallest lease duration will be used when performing the scheduling for the next refresh.
```
raffaelespazzoli, line 406: why do you say: `if no refreshPeriod is specified`? this should always be applied regardless of how the refreshPeriod is calcutaled.

readme.md
```diff
@@ -202,10 +203,24 @@ If you don't have a Vault instance available for testing, deploy one with these
 
 ```shell
 helm repo add hashicorp https://helm.releases.hashicorp.com
-oc new-project vault
-oc adm policy add-role-to-user admin -z vault -n vault
 export cluster_base_domain=$(oc get dns cluster -o jsonpath='{.spec.baseDomain}')
-helm upgrade vault hashicorp/vault -i --create-namespace -n vault --atomic -f ./config/local-development/vault-values.yaml --set server.route.host=vault-vault.apps.${cluster_base_domain}
+envsubst < ./config/local-development/vault-values.yaml > /tmp/values
```
raffaelespazzoli, line 207: this is wrong, why are you adding this back?

raffaelespazzoli (changes requested):
I left some comments.

controllers/vaultsecret_controller.go
trevorbox: Not sure I understand. I only want update events when the spec changes. I don't want an update if the status field changes, for example. This is not the right way to do it?

controllers/vaultsecret_controller.go
trevorbox: yeah I probably don't need the Owns anymore. ill test removing it.

docs/api.md
trevorbox: refreshPeriod is an override and if set will always resync after the interval is passed

readme.md
trevorbox: I was confused why it was removed... how are you unsealing vault without these instructions?

config/local-development/tilt/add-vault-environment-variables.yaml
trevorbox: only when using path: /spec/template/spec/containers/0/env
does the tilt up command work for me.
What version of kustomize are you using?
Here mine...
[tbox@fedora vault-config-operator]$ kustomize version
{Version:kustomize/v4.4.1 GitCommit:b2d65ddc98e09187a8e38adc27c30bab078c1dbf BuildDate:2021-11-11T23:36:27Z GoOs:linux GoArch:amd64}

controllers/vaultsecret_controller.go
raffaelespazzoli: go it sorry, may bad

docs/api.md
raffaelespazzoli: agreed. I was referring to the comment on the refreshThreshold documentation. `refreshThreshold` should apply to `refreshPeriod` too.

readme.md
raffaelespazzoli: bunch of sidecars, I'm proud of the solution I found.

docs/api.md
trevorbox: `refreshThreshold` is really only needed for dynamic vault secrets. I think it would be confusing to apply a threshold percentage to a `refreshPeriod`, especially when the default `refreshThreshold` is 90%.

readme.md
trevorbox: fixed readme

config/local-development/tilt/add-vault-environment-variables.yaml
trevorbox: Seems like we need to fix this in a separate PR.

config/local-development/tilt/replace-image.yaml
trevorbox: fix this in separate PR

PR #25, opened 2022-02-08
Bugfix : RandomSecret is not refreshed with refreshPeriod set

pbriet:
Hello,

I believe there is a bug here.

Do not generate password (return) if : 

* Password has already been generated :  `instance.Status.LastVaultSecretUpdate != nil`
* AND
  * There is no refresh period  `instance.Spec.RefreshPeriod == nil`  (BUGFIX : was the opposite)
  * OR there is a refresh period which is elapsed : `(instance.Spec.RefreshPeriod != nil && !instance.Status.LastVaultSecretUpdate.Add(instance.Spec.RefreshPeriod.Duration).Before(time.Now())))`


raffaelespazzoli:
LGTM

PR #35, opened 2022-03-03
Added Quay Secrets Engine

sabre1041:
Added support for the [Added Quay Secrets Engine](https://github.com/redhat-cop/vault-plugin-secrets-quay)


PROJECT
```diff
@@ -213,4 +213,43 @@ resources:
...
+    defaulting: true
+    validation: true
+    webhookVersion: v1
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: redhat.io
+  group: redhatcop
+  kind: QuaySecretEngineStaticRole
```
raffaelespazzoli, line 248: what is a static role? Can you explain this concept, other engines don't need it.

api/v1alpha1/quaysecretengineconfig_webhook.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+
+//+kubebuilder:webhook:path=/mutate-redhatcop-redhat-io-v1alpha1-quaysecretengineconfig,mutating=true,failurePolicy=fail,sideEffects=None,groups=redhatcop.redhat.io,resources=quaysecretengineconfigs,verbs=create;update,versions=v1alpha1,name=mquaysecretengineconfig.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Defaulter = &QuaySecretEngineConfig{}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (r *QuaySecretEngineConfig) Default() {
```
raffaelespazzoli, line 44: if you are not setting any default, are you sure you need this?

api/v1alpha1/quaysecretenginerole_webhook.go
```diff
@@ -0,0 +1,68 @@
...
+		For(r).
+		Complete()
+}
+
+//+kubebuilder:webhook:path=/mutate-redhatcop-redhat-io-v1alpha1-quaysecretenginerole,mutating=true,failurePolicy=fail,sideEffects=None,groups=redhatcop.redhat.io,resources=quaysecretengineroles,verbs=create;update,versions=v1alpha1,name=mquaysecretenginerole.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Defaulter = &QuaySecretEngineRole{}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (r *QuaySecretEngineRole) Default() {
```
raffaelespazzoli, line 40: again not sure this is needed

api/v1alpha1/quaysecretenginerole_webhook.go
```diff
@@ -0,0 +1,68 @@
...
+	quaysecretenginerolelog.Info("default", "name", r.Name)
+
+}
+
+//+kubebuilder:webhook:path=/validate-redhatcop-redhat-io-v1alpha1-quaysecretenginerole,mutating=false,failurePolicy=fail,sideEffects=None,groups=redhatcop.redhat.io,resources=quaysecretengineroles,verbs=create;update,versions=v1alpha1,name=vquaysecretenginerole.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &QuaySecretEngineRole{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *QuaySecretEngineRole) ValidateCreate() error {
```
raffaelespazzoli, line 50: also this does not seem to be needed

api/v1alpha1/quaysecretenginerole_webhook.go
```diff
@@ -0,0 +1,68 @@
...
+	quaysecretenginerolelog.Info("validate create", "name", r.Name)
+
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *QuaySecretEngineRole) ValidateUpdate(old runtime.Object) error {
```
raffaelespazzoli, line 57: you probably need to at least make sure that path is not changed.

api/v1alpha1/quaysecretenginestaticrole_types.go
```diff
@@ -0,0 +1,129 @@
+/*
```
raffaelespazzoli, line 1: so a static tole is a role with no expiration? Why can't you handle that with an special TTL? maybe negative of maxint? This would save us a an API.

controllers/quaysecretengineconfig_controller.go
```diff
@@ -0,0 +1,228 @@
...
+	ctx = context.WithValue(ctx, "vaultClient", vaultClient)
+	vaultResource := vaultresourcecontroller.NewVaultResource(&r.ReconcilerBase, instance)
+
+	return vaultResource.Reconcile(ctx, instance)
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *QuaySecretEngineConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	// this will filter routes that have the annotation and on update only if the annotation is changed.
```
raffaelespazzoli, line 86: check the comment

docs/secret-engines.md
```diff
@@ -194,6 +197,130 @@ to read a new credential from this role, execute the following:
...
+
+This CR is roughly equivalent to this Vault CLI command:
+
+```shell
+vault write [namespace/]quay/demo/config url=https://<QUAY_HOST> token=<token>
+```
+
+## QuayHubSecretEngineRole
+
+The `QuayHubSecretEngineRole` CRD allows a user to create a Quay secret engine role. A role allows for the creation of a narrowly scoped Quay Robot account within an organization.
```
raffaelespazzoli, line 240: the `Hub` necessary on the CRD name? It does not match well with the engine config.

docs/secret-engines.md
```diff
@@ -194,6 +197,130 @@ to read a new credential from this role, execute the following:
...
+vault write [namespace/]quay/demo/roles/repo-manager namespaceName=myorg createRepositories=true defaultPermission=write TTL=8760h
+```
+
+to read a new credential from this role, execute the following:
+
+```shell
+vault read quay/demo/creds/repo-manager
+```
+
+## QuayHubSecretEngineRole
```
raffaelespazzoli, line 283: I don't see the difference with the one above....

raffaelespazzoli (changes requested):
please see the comments. Main point: please see if it is possible to use only 2 CRs.

api/v1alpha1/quaysecretenginestaticrole_types.go
sabre1041: no unfortunately Vault does not support infinite TTL

docs/secret-engines.md
sabre1041: Bad docs. I updated the content to correct.

docs/secret-engines.md
sabre1041: Bad docs. I updated the content to correct.

PROJECT
sabre1041: This secrets engine is modeled in a similar way as the Database secrets engine. You can read [here](https://www.vaultproject.io/docs/secrets/databases#static-roles) on the concepts of Static Roles

api/v1alpha1/quaysecretenginerole_webhook.go
sabre1041: Updated

controllers/quaysecretengineconfig_controller.go
sabre1041: Can you elaborate on this?

api/v1alpha1/quaysecretengineconfig_webhook.go
sabre1041: I see some Secrets engines setting finalizers, while others do not. do you have any advice around this topic?

PROJECT
raffaelespazzoli: ok, now I see, but does it need to be a different API. Could it just be a flag?

api/v1alpha1/quaysecretengineconfig_webhook.go
raffaelespazzoli: for some secret engine the config object cannot be delete as it's 1 to 1 with the engine mount. so we have decided not to do anything on the finalize. but you should have finalizes if this engine support an 1 to many relation. It should ve definitely needed for role as they are always one to many.

api/v1alpha1/quaysecretenginestaticrole_types.go
raffaelespazzoli: see my initial comment about a flag for distinguish static and dynamic role.

PROJECT
sabre1041: Right now it could. however, the next feature that will be added in the plugin is the ability to define a rotation period. Once again, could be added to a single Role CR, but could make things even less clear

api/v1alpha1/quaysecretengineconfig_webhook.go
sabre1041: You should be able to define multiple configs at different paths. so i can look to add in support for the finalizer and will update

sabre1041: Not sure if the plugin falls into that or not. i believe we have a 1:1 with the path (ie you cant have multiple configs on the same path)

PROJECT
raffaelespazzoli: so we keep this as a separate API

PR #38, opened 2022-03-13
Helmchart integration test

### LINKED ISSUE #37: Fix the Helm chart and add a chart deployment test
The Helm chart can have an integration test deployment to a kind cluster. This should test that the latest tag works along with metrics collection from a prometheus instance.
Additionally, the chart needs a number of fixes including the selector in the Deployment/Service and working cert-manager certificates creation for metrics and the webhook.

trevorbox:
Closes #37


.github/workflows/pr.yaml
```diff
@@ -8,7 +8,8 @@ on:
 jobs:
   shared-operator-workflow:
     name: shared-operator-workflow
-    uses: redhat-cop/github-workflows-operators/.github/workflows/pr-operator.yml@v1.0.1
+    uses: trevorbox/github-workflows-operators/.github/workflows/pr-operator.yml@helmchart-test
```
trevorbox, line 11: Waiting on https://github.com/redhat-cop/github-workflows-operators/pull/9

raffaelespazzoli:
please see if we can use this approach https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#tlsconfig

config/prometheus/monitor.yaml
```diff
@@ -14,7 +14,11 @@ spec:
       port: https
       scheme: https
       tlsConfig:
-        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+        ca: 
+          secret:
+            key: 'tls.crt'
+            name: vault-config-operator-certs
+            optional: false
```
trevorbox, line 21: @raffaelespazzoli for some reason I didn't need to give the prometheus serviceaccount permission (update the Role in the rbac config section) to read Secrets in the namespace. Tested successfully using kind and in OCP after a bundle deploy.

trevorbox: Sounds like the operator has access to the Secrets, prometheus itself doesn't need rbac to Secrets.

trevorbox:
@raffaelespazzoli this is ready for review

raffaelespazzoli:
should fix #44

PR #45, opened 2022-04-10
locks go version to 1.16.

### LINKED ISSUE #44: If use of this operator is compatible with non-OpenShift Kubernetes, can that please be documented?
(I'm writing from memory and can add more concrete information as needed)

When reviewing content such as:
- https://github.com/redhat-cop/vault-config-operator#deploying-the-operator
- https://cloud.redhat.com/blog/configuring-vault-for-kubernetes-an-operator-based-approach
- https://youtu.be/Bce_0qa6ias

...I got really interested in the use of this operator - looks super cool! Unfortunately I am not an Openshift user, and in my casual attempts to install it in DIY fashion, I ran into headaches about my OLM catalog since I don't have the preloaded ones, just the single community catalog that points at OperatorHub. When I stopped trying to use OLM for the install, I then had hiccups about the image registry when trying to use the Helm chart more directly, since I don't have a pullSecret for any of Red Hat's image repositories on Quay.

Is this operator supported for use on vanilla Kube, or perhaps able to be used without any expectation of support if I had a bit more information? Or is it always expected to run in the context of OpenShift specifically?

raffaelespazzoli:
fixes #44 
Signed-off-by: raffaelespazzoli <raffaele.spazzoli@gmail.com>

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -8,7 +8,7 @@ on:
 jobs:
   shared-operator-workflow:
     name: shared-operator-workflow
-    uses: redhat-cop/github-workflows-operators/.github/workflows/pr-operator.yml@main
+    uses: redhat-cop/github-workflows-operators/.github/workflows/pr-operator.yml@v1.0.3
     with: 
       RUN_UNIT_TESTS: true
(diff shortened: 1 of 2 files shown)


sabre1041 (approved):
LGTM

PR #53, opened 2022-04-28
refactor kind setup and break out into deploy-ingress and deploy-vaul…

trevorbox:
…t in Makefile


trevorbox:
Closes #54

.github/workflows/pr.yaml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   shared-operator-workflow:
     name: shared-operator-workflow
-    uses: redhat-cop/github-workflows-operators/.github/workflows/pr-operator.yml@v1.0.3
+    uses: trevorbox/github-workflows-operators/.github/workflows/pr-operator.yml@default-helmchart-version
```
raffaelespazzoli, line 11: this is probably not correct, do you need me to create a new tag on github-workflows-operators?

raffaelespazzoli (changes requested):

.github/workflows/pr.yaml
trevorbox: its correct now.

PR #87, opened 2022-12-08
Fix DatabaseSecretEngineConfig so that it works with v2 RandomSecret.

ameyp:
Also add some tests for DatabaseSecretEngine and DatabaseSecretEngineConfig.

### DIFF
--- a/api/v1alpha1/databasesecretengineconfig_types.go
+++ b/api/v1alpha1/databasesecretengineconfig_types.go
@@ -45,6 +45,7 @@ type DatabaseSecretEngineConfigSpec struct {
 	// +kubebuilder:validation:Required
 	Path vaultutils.Path `json:"path,omitempty"`
 
+	// +kubebuilder:validation:Required
 	DBSEConfig `json:",inline"`
 
 	// RootCredentials specifies how to retrieve the credentials for this DatabaseEngine connection.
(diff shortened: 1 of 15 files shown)


raffaelespazzoli:
something wrong with this PR as it seems to contain the work for static roles which has been merged yesterday. Can you clean it up?

api/v1alpha1/databasesecretengineconfig_types.go
```diff
@@ -151,7 +159,6 @@ type DBSEConfig struct {
 
 	// VerifyConnection Specifies if the connection is verified during initial configuration. Defaults to true.
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default=true
```
raffaelespazzoli, line 154: please leave this here, otherwise the default will be false.

api/v1alpha1/databasesecretenginerole_types.go
```diff
@@ -44,6 +44,7 @@ type DatabaseSecretEngineRoleSpec struct {
 	// +kubebuilder:validation:Required
 	Path vaultutils.Path `json:"path,omitempty"`
 
+	// +kubebuilder:validation:Required
```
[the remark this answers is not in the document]
ameyp: Ok, removed

api/v1alpha1/databasesecretengineconfig_types.go
ameyp: It doesn't work if I leave it there. Even with `verifyConnection: false` in a supplied YAML, the custom resource created by the test has `verifyConnection: true` in it, and the controller sends `verify_connection: true` in the payload to Vault. If I `kubectl apply` the YAML file, the created custom resource has `verifyConnection: false` in it, which leads me to believe that there's a bug somewhere else in the existing integration test code, but I couldn't find it.

Without the `default=true`, the YAML value is respected. Also FWIW, if you omit the argument altogether, Vault assumes it to be `true`. RabbitMQ secret engine's config works the same way: https://github.com/redhat-cop/vault-config-operator/blob/14b154211ca0f4515715c92d334c6f64b656e4da/api/v1alpha1/rabbitmqsecretengineconfig_types.go#L67-L69

PR #94, opened 2022-12-31
Add volumes and volumeMounts customization to Helm chart.

ameyp:
Additional volumes and volume mounts can be injected using OLM, but the Helm chart does not provide the same functionality. This makes it impossible to specify custom certificates for vault when the operator is installed using the Helm chart.


ameyp:
@raffaelespazzoli Could you take a look at this PR? I need this feature for automated tests in my cluster

PR #99, opened 2023-01-02
Specify default values for optional usernameKey and passwordKey.

### LINKED ISSUE #98: Secret engines with rootCredentials don't work without explicit usernameKey and passwordKey
I applied the following JSON:

```
{
   "apiVersion": "redhatcop.redhat.io/v1alpha1",
   "kind": "DatabaseSecretEngineConfig",
   "metadata": {
      "name": "postgresql",
      "namespace": "vault"
   },
   "spec": {
      "allowedRoles": [
         "*"
      ],
      "authentication": {
         "path": "kubernetes",
         "role": "vault-admin"
      },
      "connectionURL": "postgresql://{{username}}:{{password}}@postgresql.example.com/mydb?sslmode=require",
      "path": "database/postgresql",
      "pluginName": "postgresql-database-plugin",
      "rootCredentials": {
         "secret": {
            "name": "my-database-secret"
         }
      }
   }
}

```

and reconcile failed with:

```
error creating database object: error verifying connection: failed to connect to `host=postgresql.example.com user= database=mydb`: server error (FATAL: no PostgreSQL user name specified in startup packet (SQLSTATE 28000))
```

The secret exists and is of type `basic-auth` with `username` and `password` keys.

ameyp:
Without these, at least DatabaseSecretEngineConfig does not work because the operator does not specify a username or password in the request made to Vault for configuring the engine.

Fixes #98

### DIFF
--- a/api/v1alpha1/utils/commons.go
+++ b/api/v1alpha1/utils/commons.go
@@ -190,10 +190,12 @@ type RootCredentialConfig struct {
 
 	// PasswordKey key to be used when retrieving the password, required with VaultSecrets and Kubernetes secrets, ignored with RandomSecret
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="password"
 	PasswordKey string `json:"passwordKey,omitempty"`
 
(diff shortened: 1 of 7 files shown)


raffaelespazzoli:
I like this one, thanks for thinking about it.

PR #120, opened 2023-02-06
fixes #95

### LINKED ISSUE #95: [RFE] Ability o configure multiple Vault instances with the same operator
Hi !
I've seen in this section https://github.com/redhat-cop/vault-config-operator/blob/842f4c2dbe01ee69b56558ce169e3a7aeb97a1ba/docs/auth-section.md how to authenticate to the remote Vault instance. However it seems it's not possible to have two CRs configuring two different Vault instances. Would it be possible to add an "address" field to authentication section ?

raffaelespazzoli:
fixes #95


config/local-development/vault-values.yaml
```diff
@@ -5,9 +5,24 @@ global:
...
+  image:
+    repository: "docker.io/hashicorp/vault-csi-provider"
+    tag: "1.2.1"
+    pullPolicy: IfNotPresent     
+
 server:
   image:
-    tag: "1.12.0"
+    tag: "1.12.2"
+    repository: docker.io/hashicorp/vault
```
trevorbox, line 25: @raffaelespazzoli any reason for using the docker image vs https://github.com/hashicorp/vault-helm/blob/main/values.openshift.yaml#L17-L18

raffaelespazzoli: actually, I'd do the change and recommit

PR #122, opened 2023-02-20
Fix ReadME: step to get secret token for kube auth

zengwuxin:


readme.md
```diff
@@ -338,7 +338,7 @@ export accessor=$(vault read -tls-skip-verify -format json sys/auth | jq -r '.da
 verify that kube authentication works:
 
 ```sh
-export token=$(oc create token default -n vault-admin)
+token=$(oc get secret DEFAULT_SERVICE_ACCOUNT_SECRET_NAME -n vault-admin -o jsonpath='{.data.token}' | base64 -d)
```
raffaelespazzoli, line 341: why would you like to make that change?

raffaelespazzoli (changes requested):

readme.md
zengwuxin: `oc create token default -n vault-admin` command doesn't give the service account token to use in the next step, so I tested with this command `oc get secret DEFAULT_SERVICE_ACCOUNT_SECRET_NAME -n vault-admin -o jsonpath='{.data.token}' | base64 -d` to get the right token, where DEFAULT_SERVICE_ACCOUNT_SECRET_NAME is the default service account token secret in the vault-admin namespace.

zengwuxin:
Updating oc client to 4.11.x allowed me to use the `oc create token default` command . Closing the PR.

PR #130, opened 2023-02-27
Fixes#126

### LINKED ISSUE #126: Resource deletion stalls and is not propagated to Vault
Hi,

Using a Vault with kv-v2 enabled, creating a `RandomSecret` does properly create the entry into the Vault kv store.

However, removing the `RandomSecret` resource stalls after acknowledgment, and the Vault entry is not removed.

This is due to the fact that the finalizer key on the `RandomSecret` object is never removed - so Kubernetes awaits forever.
Manually removing the finalizer obviously unlocks the Kubernetes situation, but the Vault entry is left alone and still exists.

As far as I can tell, it does not look normal as the removal should be performed onto Vault as well.

By tracing calls to the `RandomSecretReconciler` controller, it turns out that the controller is not invoked upon the CR update on deletion: per https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/#how-finalizers-work, the deletion operation adds the deletionTimestamp (updating the CR object), then the Kube control plane awaits that the finalizer list is empty before actually removing the object.

Updating the `RandomSecret` CR is also entirely ignored - controller is not invoked, and no change is performed in Vault.

Note that this actually happens for any Custom Resource - removing an entire namespace containing multiple CRs from this operator hangs, Kubernetes reporting them as awaiting for finalizers.

This situation happens running a Kubernetes cluster 1.24.8 (GKE) with `vault-config-operator` version 0.8.10 deployed using Helm.

Is there anything I missed during deployment or configuration?
Thanks in advance,
Regards

trevorbox:
fixes #126


config/crd/bases/redhatcop.redhat.io_authenginemounts.yaml
```diff
@@ -170,6 +170,7 @@ spec:
                               TODO: Add other useful fields. apiVersion, kind, uid?'
                             type: string
                         type: object
+                        x-kubernetes-map-type: atomic
```
raffaelespazzoli, line 173: what are all of these?

config/crd/bases/redhatcop.redhat.io_vaultsecrets.yaml
```diff
@@ -116,6 +116,63 @@ spec:
                           type: object
                           x-kubernetes-map-type: atomic
                       type: object
+                    connection:
```
raffaelespazzoli, line 119: why was this missing in vaultsecret?

controllers/vaultsecret_controller.go
```diff
@@ -93,6 +93,24 @@ func (r *VaultSecretReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	ctx = context.WithValue(ctx, "kubeClient", r.GetClient())
 	ctx = context.WithValue(ctx, "restConfig", r.GetRestConfig())
 
+	if util.IsBeingDeleted(instance) {
```
raffaelespazzoli, line 96: so we had an issue with finalizers also in vault secrets?

trevorbox: yes - tested both randomsecret and vaultsecret and found them both to not delete correctly

Issue #139, opened 2023-04-17
Please consider reusing vault tokens


ameyp:
The current approach of logging in to Vault and getting a fresh token during each reconcile wastes CPU and network bandwidth. I understand that it's set up this way to prevent attacks originating from leaking tokens, but caching the token is worth that risk IMO.

The problem is even worse with `VaultSecret`, where a new token is created for each `SecretDefinition` inside the custom resource. If I have a `VaultSecret` that reads from 10 different secrets in vault using the same role, the operator logs in 10 times and creates 10 different leases. I don't know how often a reconcile is run, so let's assume it runs every 5 seconds. That means that in a minute, for that one `VaultSecret`, the operator creates 120 new leases. If I have 10 `VaultSecret` resources with 10 secret definitions each, that's 1200 new leases every minute. Even if I set the token TTL to 1 minute, Vault will have to clean up 1200 leases every minute, and depending on how much CPU is assigned to the Vault pod, that might not be feasible.

Even the new [vault-secrets-operator](https://github.com/hashicorp/vault-secrets-operator) seems to be reusing tokens. I haven't looked at its code, but users are expected to define `VaultAuth` and `VaultConnection` resources and use them for each secret CR, which leads me to believe that they must be reusing tokens. The `external-secrets` operator also added an [experimental feature](https://github.com/external-secrets/external-secrets/pull/1537) recently for reusing vault tokens.

PR #142, opened 2023-04-21
Cache vault clients.

ameyp:
Fixes https://github.com/redhat-cop/vault-config-operator/issues/139

### DIFF
--- a/api/v1alpha1/utils/commons.go
+++ b/api/v1alpha1/utils/commons.go
@@ -19,8 +19,11 @@ package utils
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
+	"sync"
 
+	"github.com/go-logr/logr"
 	vault "github.com/hashicorp/vault/api"
 	authv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -30,6 +33,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
+
+var vaultClientCache = VaultClientCache{
+	clients: make(map[string]*vault.Client),
+}
 
 // +kubebuilder:object:generate=true
 // +kubebuilder:validation:Pattern:=`^(?:/?[\w;:@&=\$-\.\+]*)+/?`
@@ -74,6 +81,11 @@ type VaultConnection struct {
 	MaxRetries *int `json:"maxRetries,omitempty"`
 }
 
+type VaultClientCache struct {
+	clients map[string]*vault.Client
+	mutex sync.Mutex
+}
+
 // +kubebuilder:object:generate=true
 type TLSConfig struct {
 	// Cacert Path to a PEM-encoded CA certificate file on the local disk. This file is used to verify the Vault server's SSL certificate. This environment variable takes precedence over a cert passed via the secret.
@@ -91,6 +103,24 @@ type TLSConfig struct {
 	// TLSServerName Name to use as the SNI host when connecting via TLS.
 	// +kubebuilder:validation:Optional
 	TLSServerName *string `json:"tlsServerName,omitempty"`
+}
+
+func (cache *VaultClientCache) Get(kc *KubeAuthConfiguration) *vault.Client {
+	cache.mutex.Lock()
+	defer cache.mutex.Unlock()
+	return cache.clients[kc.getCacheKey()]
+}
+
+func (cache *VaultClientCache) Put(kc *KubeAuthConfiguration, client *vault.Client) {
+	cache.mutex.Lock()
+	defer cache.mutex.Unlock()
+	cache.clients[kc.getCacheKey()] = client
+}
+
+func (cache *VaultClientCache) Delete(kc *KubeAuthConfiguration) {
+	cache.mutex.Lock()
+	defer cache.mutex.Unlock()
+	delete(cache.clients, kc.getCacheKey())
 }
 
 func (vc *VaultConnection) getConnectionConfig(context context.Context, kubeNamespace string) (*vault.Config, error) {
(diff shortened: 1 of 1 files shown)


api/v1alpha1/utils/commons.go
```diff
@@ -74,6 +81,11 @@ type VaultConnection struct {
 	MaxRetries *int `json:"maxRetries,omitempty"`
 }
 
+type VaultClientCache struct {
+	clients map[string]*vault.Client
```
raffaelespazzoli, line 85: let's use a sync.Map[1] here so we don't have to worry about the mutex
[1]: https://pkg.go.dev/sync#Map

api/v1alpha1/utils/commons.go
```diff
@@ -153,18 +183,30 @@ func (kc *KubeAuthConfiguration) GetServiceAccountName() string {
 	return kc.ServiceAccount.Name
 }
 
+func (kc *KubeAuthConfiguration) getCacheKey() string {
+	return fmt.Sprintf("%s:%s:%s:%s", kc.ServiceAccount.Name, kc.Path, kc.Role, kc.Namespace)
```
raffaelespazzoli, line 187: the kube namespace needs to be in key otherwise you could have two authentication config in different namespace obtaining the same vault account

api/v1alpha1/utils/commons.go
```diff
@@ -153,18 +183,30 @@ func (kc *KubeAuthConfiguration) GetServiceAccountName() string {
...
+}
+
 func (kc *KubeAuthConfiguration) GetVaultClient(context context.Context, kubeNamespace string) (*vault.Client, error) {
 	log := log.FromContext(context)
+
+	vaultClient := vaultClientCache.Get(kc)
+	if vaultClient != nil {
+		log.Info("Returning cached client")
```
raffaelespazzoli, line 195: please turn this to debug

api/v1alpha1/utils/commons.go
```diff
@@ -240,9 +283,39 @@ func (kc *KubeAuthConfiguration) createVaultClient(context context.Context, jwt
 
 	client.SetToken(secret.Auth.ClientToken)
 
+	go kc.startLifetimeWatcher(client, secret, log)
+
 	return client, nil
 }
 
+func (kc *KubeAuthConfiguration) startLifetimeWatcher(client *vault.Client, secret *vault.Secret, log logr.Logger) {
```
raffaelespazzoli, line 291: can you explain why this is needed?

api/v1alpha1/utils/commons.go
```diff
@@ -153,18 +183,30 @@ func (kc *KubeAuthConfiguration) GetServiceAccountName() string {
...
 }
 
+func (kc *KubeAuthConfiguration) getCacheKey() string {
+	return fmt.Sprintf("%s:%s:%s:%s", kc.ServiceAccount.Name, kc.Path, kc.Role, kc.Namespace)
+}
+
 func (kc *KubeAuthConfiguration) GetVaultClient(context context.Context, kubeNamespace string) (*vault.Client, error) {
 	log := log.FromContext(context)
+
+	vaultClient := vaultClientCache.Get(kc)
```
raffaelespazzoli, line 193: add a check that the jwt token isn't expired, if it is remove it from the map and continue with creating a new one

raffaelespazzoli (changes requested):
sorry for the delay, here are my comments.

api/v1alpha1/utils/commons.go
ameyp: I thought `kc.Namespace` was the kube namespace, is it not?

api/v1alpha1/utils/commons.go
ameyp: If the TTL for the token is less than the maxTTL, the lifetime watcher renews the token until it reaches the maxTTL. Added a comment

api/v1alpha1/utils/commons.go
ameyp: Done (I think, please verify)

api/v1alpha1/utils/commons.go
ameyp: Thank you, done

api/v1alpha1/utils/commons.go
raffaelespazzoli: no, that is the vault namespace.

erlisb:
@ameyp is there any more updates/changes this ? 
We would like to have this merged.

ameyp:
> the way the cache key is calculated does not ensure uniqueness. That is why I have not accepted this PR. There is a comment about that in the PR review. @ameyp @erlisb

Apologies, I forgot about that comment. I've updated the PR.

raffaelespazzoli:
@ameyp you change looks good. I am still afraid about the possible consequences of this PR. Can we put this feature behind a flag? let' create an  env variable called CACHE_VAULT_TOKEN=[true|false]. And enable caching only when requested. Then this behavior is controllable at deploy time.
@ameyp do you mind making this change?

ameyp:
> @ameyp you change looks good. I am still afraid about the possible consequences of this PR. Can we put this feature behind a flag? let' create an env variable called CACHE_VAULT_TOKEN=[true|false]. And enable caching only when requested. Then this behavior is controllable at deploy time. @ameyp do you mind making this change?

Made that change, PTAL!

raffaelespazzoli:
LGTM

erlisb:
@ameyp I think a small section in the documentation regarding this addition, how to enable/disable it, would be helpful.
Currently, there is nothing mentioned about it in the docu, expect for this PR.

Thank you.

ameyp:
I agree. @raffaelespazzoli What would be a good place to put this?

raffaelespazzoli:
@ameyp you can add a section after here: https://github.com/redhat-cop/vault-config-operator#initializing-the-connection-to-vault, explain that parameter.

ameyp:
Thanks, created https://github.com/redhat-cop/vault-config-operator/pull/179

PR #144, opened 2023-05-12
Fixes #143

### LINKED ISSUE #143: Support for automatically resolving the <mount accessor> in Policy
Creating a templated Policy using a mount accessor currently requires that this mount accessor is resolved to its value, as described in https://github.com/redhat-cop/vault-config-operator/blob/main/docs/policy-management.md#policy

This makes it hard to automate Vault configuration using a GitOps or similar process, since it requires querying Vault first (possibly using an API request to `/sys/auth`, see https://developer.hashicorp.com/vault/api-docs/system/auth#list-auth-methods), generate the proper `Policy` content with the resolved mount accessor, then push the k8s Custom Resource so that it's created onto Vault.

Being able to specify the mount path instead and leave the resolution step to the operator would be ideal.
This is following the same line as in https://github.com/banzaicloud/bank-vaults/issues/1533

Any idea how to do this, or existing thoughts about introducing such resolution step into the operator?

davoustp:
This implementation aims at automatically resolving Vault auth engine accessors in templated Policies.

This allows to use such templated Policies in a declarative deployment approach without requiring any manual or scripted retrieval of the accessor values (who said gitops?).

Placeholders formatted as `${auth/<auth engine mount>/@accessor}` are replaced with the corresponding value.
This is of course a proposal, and open to discussion!

BTW, it looks like the integration tests would directly benefit from this, as the Kubernetes accessor is retrieved during deployment (see [Makefile@124](https://github.com/redhat-cop/vault-config-operator/blob/main/Makefile#L124)) and used in the test themselves (see [controllers/suite_integration_test.go](https://github.com/redhat-cop/vault-config-operator/blob/main/controllers/suite_integration_test.go#L83) , [controllers/vaultsecret_controller_test.go](https://github.com/redhat-cop/vault-config-operator/blob/main/controllers/vaultsecret_controller_test.go) and [controllers/vaultsecret_controller_v2_test.go](https://github.com/redhat-cop/vault-config-operator/blob/main/controllers/vaultsecret_controller_v2_test.go) ).

Let me know if you want me to have a cut at this as well.

### DIFF
--- a/api/v1alpha1/policy_types.go
+++ b/api/v1alpha1/policy_types.go
@@ -18,11 +18,16 @@ package v1alpha1
 
 import (
 	"context"
+	"errors"
 	"reflect"
+	"regexp"
+	"strings"
 
+	vault "github.com/hashicorp/vault/api"
 	vaultutils "github.com/redhat-cop/vault-config-operator/api/v1alpha1/utils"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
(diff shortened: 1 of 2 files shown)


api/v1alpha1/policy_types.go
```diff
@@ -60,6 +65,35 @@ func (d *Policy) IsInitialized() bool {
...
+	// Retrieves the list of auth engines to get their accessors
+	// Kinda duplicates logic found in VaultEngineObject.retrieveAccessor
+	vaultClient := context.Value("vaultClient").(*vault.Client)
+	secret, err := vaultClient.Logical().Read("sys/auth")
+	if err != nil {
+		// Log but ignore the error: do not resolve placeholders
+		log.Error(err, "could not resolve auth engine accessor(s) in policy rule - unable to retrieve auth engines at", "path", "sys/auth")
+		return nil
+	}
+	if secret == nil {
```
raffaelespazzoli, line 85: why would this result in an error if no auth are found?

api/v1alpha1/policy_types.go
```diff
@@ -60,6 +65,35 @@ func (d *Policy) IsInitialized() bool {
...
+		return nil
+	}
+
+	log := log.FromContext(context)
+
+	// Retrieves the list of auth engines to get their accessors
+	// Kinda duplicates logic found in VaultEngineObject.retrieveAccessor
+	vaultClient := context.Value("vaultClient").(*vault.Client)
+	secret, err := vaultClient.Logical().Read("sys/auth")
+	if err != nil {
```
raffaelespazzoli, line 80: I understand why you mapped this to an error, but should we? This would mean that now everyone who is trying to create a policy must also have access to the sys/auth path. Maybe that is not desirable. What about just logging the problem and continuing with an error and of course without replacement.
Actually, even better can we check if there is anything to replace and if there is then we return an error?

raffaelespazzoli (changes requested):
I like this improvement, please take a look at my comments.

api/v1alpha1/policy_types.go
davoustp: Well, the Token auth method is automatically enabled (see https://developer.hashicorp.com/vault/docs/auth/token) so no error and no auth method means something is wrong, unless I'm mistaken?

api/v1alpha1/policy_types.go
davoustp: I'm not sure I fully understand the question...

> This would mean that now everyone who is trying to create a policy must also have access to the sys/auth path.

Actually, no: if there is no `${...}` placeholder in the policy rules (checked at line 69), then line 70 will exit without further processing and no error - no change compared to existing code.

In case there is at least one `${...}` placeholder but an error is returned by `sys/auth`, then the error is logged at line 80 but no error is returned (same behavior as existing code, no replacement done nor error raised).

> What about just logging the problem and continuing with an error and of course without replacement.

Yes, I believe that this is the behavior that is implemented, the only case when an error is raised is when there is at least one placeholder to replace but sys/auth returned nothing, without error.
Unless I've missed sthg, of course ;-)

> Actually, even better can we check if there is anything to replace and if there is then we return an error?

This check is done at the very beginning of the method, see above.

davoustp: @raffaelespazzoli any news or anything I can provide to help getting that through?
Thx

PR #147, opened 2023-05-22
Change condition type back to apis.ReconcileSuccess to fix deletion

npurdy-tyro:
It looks like a change was made that changed the status for success from a type of "ReconcileSuccess" to "ReconcileSuccessful". This means that the code (manageCleanUpLogic) that performs a delete wasn't recognizing that the configuration was successfully applied and therefore wouldn't delete configuration from vault.

I have changed it back to use the redhat operator utils "apis.ReconcileSuccess" so that its applied consistently throughout the project. Deletion works afterwards.


npurdy-tyro:
Looks like its been fixed by https://github.com/redhat-cop/vault-config-operator/pull/150

PR #149, opened 2023-05-24
fixes#148

davoustp:
Simplified Makefile, integration tests around the k8s auth engine accessor.


raffaelespazzoli:
please see #150 , that is the way to fix this issue. Thank you very much for finding it.

trevorbox:
@davoustp is it safe to remove the accessor exports in the README? for example https://github.com/redhat-cop/vault-config-operator#configure-an-kubernetes-authentication-mount-point

davoustp:
> please see #150 , that is the way to fix this issue. Thank you very much for finding it.

@raffaelespazzoli not sure that this relates to this PR, but thanks nonetheless ;-)

trevorbox:
> > @davoustp is it safe to remove the accessor exports in the README? for example https://github.com/redhat-cop/vault-config-operator#configure-an-kubernetes-authentication-mount-point
> 
> Blimey! thought I had this cleared as well. Yes, it should be safe. What's the best course, another PR probably?

That's fine then we can update it later if not now.

PR #158, opened 2023-06-04
fixes #156

### LINKED ISSUE #156: RandomSecret spec.secretFormat.inlinePasswordPolicy requires using a special secretKey name
The following `RandomSecret` definition, with a "naive" use of `inlinePasswordPolicy` is rejected by the validation webhook (see error below):

```yaml
kind: RandomSecret
metadata:
  name: test
spec:
  ...
  connection:
     ...
  authentication:
     ...
  ...
  secretKey: bootstrapPassword
  secretFormat:
    inlinePasswordPolicy: |
      length = 12
      rule "charset" {
      charset = "abcdefghijklmnopqrstuvwxyz"
      min-chars = 1
      }
      rule "charset" {
      charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
      min-chars = 1
      }
      rule "charset" {
      charset = "0123456789"
      min-chars = 1
      }
```

This results in the following error being raised by the admission webhook:
```
* <nil>: Unsupported file format; Cannot read from password: unrecognized file format suffix "".
, error: admission webhook "vrandomsecret.kb.io" denied the request: 1 error occurred:
	* <nil>: Unsupported file format; Cannot read from password: unrecognized file format suffix "".
```

This is due to the following:
* https://github.com/redhat-cop/vault-config-operator/blob/d869dc2cfff322ebf44ad3e04a3633b9563804ac/api/v1alpha1/randomsecret_types.go#L309
* the first parameter is `.Spec.SecretKey`  and is used as a file name by `hclsimple.Decode`, which tries to use it as a filename, and returns with an error if the string does not end with `.hcl` or `.json`:
https://github.com/hashicorp/hcl/blob/7208bce57fadb72db3a328ebc9aa86489cd06fce/hclsimple/hclsimple.go#L59-L72
```golang
func Decode(filename string, src []byte, ctx *hcl.EvalContext, target interface{}) error {
        var file *hcl.File
        var diags hcl.Diagnostics

        switch suffix := strings.ToLower(filepath.Ext(filename)); suffix {
        case ".hcl":
                file, diags = hclsyntax.ParseConfig(src, filename, hcl.Pos{Line: 1, Column: 1})
        case ".json":
                file, diags = json.Parse(src, filename)
        default:
                diags = diags.Append(&hcl.Diagnostic{
                        Severity: hcl.DiagError,
                        Summary:  "Unsupported file format",
                        Detail:   fmt.Sprintf("Cannot read from %s: unrecognized file format suffix %q.", filename, suffix),
```

Adding `.hcl` to the `secretKey` attribute, solves the issue:

```yaml
kind: RandomSecret
metadata:
  name: test
spec:
...
  secretKey: bootstrapPassword.hcl             # <<<<<<<<<<<<<<<<<<<<<<<<<<<<
  secretFormat:
    inlinePasswordPolicy: |
      length = 12
      rule "charset" {
      charset = "abcdefghijklmnopqrstuvwxyz"
      min-chars = 1
      }
      rule "charset" {
      charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
      min-chars = 1
      }
      rule "charset" {
      charset = "0123456789"
      min-chars = 1
      }
```

... but it's quite inelegant ;-)

raffaelespazzoli:
fixes #156 

hopefully

### DIFF
--- a/api/v1alpha1/randomsecret_types.go
+++ b/api/v1alpha1/randomsecret_types.go
@@ -306,7 +306,7 @@ func (r *RandomSecret) validateEitherPasswordPolicyReferenceOrInline() error {
 func (r *RandomSecret) validateInlinePasswordPolicyFormat() error {
 	if r.Spec.SecretFormat.InlinePasswordPolicy != "" {
 		passwordPolicyFormat := &PasswordPolicyFormat{}
-		return hclsimple.Decode(r.Spec.SecretKey, []byte(r.Spec.SecretFormat.InlinePasswordPolicy), nil, passwordPolicyFormat)
+		return hclsimple.Decode("dummy.hcl", []byte(r.Spec.SecretFormat.InlinePasswordPolicy), nil, passwordPolicyFormat)
 	}
 	return nil
 }


api/v1alpha1/randomsecret_types.go
```diff
@@ -306,7 +306,7 @@ func (r *RandomSecret) validateEitherPasswordPolicyReferenceOrInline() error {
 func (r *RandomSecret) validateInlinePasswordPolicyFormat() error {
 	if r.Spec.SecretFormat.InlinePasswordPolicy != "" {
 		passwordPolicyFormat := &PasswordPolicyFormat{}
-		return hclsimple.Decode(r.Spec.SecretKey, []byte(r.Spec.SecretFormat.InlinePasswordPolicy), nil, passwordPolicyFormat)
```
sabre1041, line 309: I'd recommend ensuring that a valid extension is present on the value retrieved from `secretKey` property. If it is not present, it can be added to sure that the requirement enforced by this method is satisfied instead of a hardcoded value

raffaelespazzoli: done

api/v1alpha1/randomsecret_types.go
```diff
@@ -306,7 +307,11 @@ func (r *RandomSecret) validateEitherPasswordPolicyReferenceOrInline() error {
 func (r *RandomSecret) validateInlinePasswordPolicyFormat() error {
 	if r.Spec.SecretFormat.InlinePasswordPolicy != "" {
 		passwordPolicyFormat := &PasswordPolicyFormat{}
-		return hclsimple.Decode("dummy.hcl", []byte(r.Spec.SecretFormat.InlinePasswordPolicy), nil, passwordPolicyFormat)
+		if strings.HasSuffix(".hcl", r.Spec.SecretKey) {
```
sabre1041, line 310: The `HasSuffix` function requires the variable being tested to be the first parameter. Swap the parameters

sabre1041 (changes requested):

sabre1041 (approved):
LGTM

PR #170, opened 2023-07-07
Fix deletion problems, add integration tests for deletion & improve the readme

trevorbox:
This should resolve several problems with deletions immediately.


trevorbox:
Fixes #171

controllers/vaultresourcecontroller/utils.go
```diff
@@ -47,7 +47,7 @@ func ManageOutcomeWithRequeue(context context.Context, r util.ReconcilerBase, ob
 	} else {
 		r.GetRecorder().Event(obj, "Warning", "ProcessingError", issue.Error())
 		condition = metav1.Condition{
-			Type:               ReconcileSuccessful,
+			Type:               ReconcileSuccessful, // TODO - this should be apis.ReconcileError?
```
trevorbox, line 50: @raffaelespazzoli should this be apis.ReconcileError?

controllers/vaultresourcecontroller/utils.go
```diff
@@ -21,7 +21,7 @@ import (
 	"time"
 
 	"github.com/redhat-cop/operator-utils/pkg/util"
-	"github.com/redhat-cop/operator-utils/pkg/util/apis"
+	"github.com/redhat-cop/operator-utils/pkg/util/apis" // TODO - are we trying to remove the apis dependency and its status completely since we are using the ReconcileSuccessful constant instead?
```
trevorbox, line 24: @raffaelespazzoli are we trying to remove the apis dependency and its status constants completely since we are using the ReconcileSuccessful constant instead? seems inconsistent

trevorbox: Goal is to remove this dependency

controllers/vaultresourcecontroller/utils.go
trevorbox: yes it should be error

PR #174, opened 2023-07-11
Fix KVv2 RandomSecret deletion and add Integration tests for deletion

trevorbox:


main.go
```diff
@@ -298,6 +298,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err = (&controllers.DatabaseSecretEngineStaticRoleReconciler{
```
raffaelespazzoli, line 301: there should not need to be a change here. Why is this needed?

raffaelespazzoli (changes requested):
please review the question

main.go
trevorbox: The webhook needs to be in the correct if statement to be controlled by the ENABLE_WEBHOOKS environment variable correctly and the ability to disable webhooks is needed when debugging. the other block was moved to put it in the same section as the other manager setups.

PR #203, opened 2023-10-10
Refactor makefile

trevorbox:
Several updates and improvements to makefile and helmchart to make testing easier. Updated configs for latest.

### DIFF
--- a/config/rbac/leader_election_role.yaml
+++ b/config/rbac/leader_election_role.yaml
@@ -29,7 +29,7 @@ rules:
   - patch
   - delete
 - apiGroups:
-  - "events.k8s.io"
+  - ""
   resources:
   - events
   verbs:
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,20 @@ CHART_REPO_URL ?= http://example.com
 HELM_REPO_DEST ?= /tmp/gh-pages
 OPERATOR_NAME ?=$(shell basename -z `pwd`)
 HELM_VERSION ?= v3.11.0
-KIND_VERSION ?= v0.17.0
-KUBECTL_VERSION ?= v1.25.3
-K8S_MAJOR_VERSION ?= 1.25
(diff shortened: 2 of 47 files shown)


config/rbac/leader_election_role.yaml
raffaelespazzoli, line 1: is this correct?

raffaelespazzoli (commented):
please see the question. that change may disable the operator in some versions of kube.

config/rbac/leader_election_role.yaml
trevorbox: This fixes the deprecated and removed api. https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25

Example warning using previous events.k8s.io apiGroup:

```sh
[tbox@fedora vault-config-operator]$ operator-sdk bundle validate ./bundle --select-optional suite=operatorframework --optional-values=k8s-version=1.25
WARN[0000] Warning: Value : (vault-config-operator.v0.0.1) csv.Spec.minKubeVersion is not informed. It is recommended you provide this information. Otherwise, it would mean that your operator project can be distributed and installed in any cluster version available, which is not necessarily the case for all projects. 
WARN[0000] Warning: Value vault-config-operator.v0.0.1: this bundle is using APIs which were deprecated and removed in v1.25. More info: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25. Migrate the API(s) for events: (["ClusterServiceVersion.Spec.InstallStrategy.StrategySpec.Permissions[0].Rules[2]"]) 
WARN[0000] Warning: Value vault-config-operator.v0.0.1: this bundle is using APIs which were deprecated and removed in v1.25. More info: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25. Migrate the API(s) for events: (["ClusterServiceVersion.Spec.InstallStrategy.StrategySpec.Permissions[0].Rules[2]"]) 
```

PR #210, opened 2023-12-04
Regenerate CRDs to fix broken name pattern

tbalzer:
During #202 (specifically cc426f8b94869fc4ce4e8e1dff6d7382752b3b70) the quoting of the validation patterns for the new `name` field was fixed, but it was missed to regenerate the CRDs, which I have done with this PR. Currently the validation does not work correctly when applying as-is from e.g. helm because of this.


tbalzer:
@raffaelespazzoli @Vikaspogu @sabre1041 Would be nice to get this fixed soon, as it's pretty trivial but breaks our deployments when trying to use CRDs from the helm release.

PR #235, opened 2024-05-10
FIX #230 CRD passwordAuthentication for PostgreSQL Database Secret Engine Configs

### LINKED ISSUE #230: password_authentication support for databaseSecretEngineConfig
Hello,

we would like to set password_authentication to "scram-sha256" through databaseSecretEngineConfig ressources but current CR and controller don't support this parameter yet.

How handle it properly ?

We can try to provide a MR in the next days

gcpcscol:
Hello,

Here's a proposal for using the passwordAuthentication option with the value scram-sha-256
Thanks

Gauthier


raffaelespazzoli:
hello, what were you trying to do with this PR? 
you cannot change the crds..., crds are generated from code and code annotations.

gcpcscol:
We noticed that CRD was missing to enable vault to encrypt the password when initializing the postgresql role.
I don't understand why the entry is missing, but adding it manually fixes the #230 issue.

PR #305, opened 2026-03-05
feat: add auth accessor resolution to IdentityTokenRole template field

dyna-stadt:
# Summary

- Extracts the `${auth/<path>/@accessor} `placeholder resolution logic from Policy into a shared `ResolveAuthAccessors` utility function in `api/v1alpha1/utils/commons.go`
- Refactors `Policy.PrepareInternalValues()` to use the new shared utility, removing duplicated code and unused imports
- Adds auth accessor resolution to` IdentityTokenRole.PrepareInternalValues()`, enabling `${auth/<path>/@accessor}` placeholders in the template field
- Updates `docs/identities.md` with documentation and an example for the new IdentityTokenRole template accessor syntax

# Motivation

The `${auth/<auth engine path>/@accessor} `placeholder syntax was previously only supported in `Policy` HCL rules. However, identity token role templates often need to reference auth engine accessors as well (e.g. `{{identity.entity.aliases.<accessor>.metadata.service_account_namespace}}`). Without this feature, users had
to manually look up and hard-code accessor values, which change across Vault instances and are not portable.

# Example

```yaml
kind: IdentityTokenRole
metadata:
  name: my-role
spec:
  authentication:
    path: kubernetes
    role: policy-admin
  key: my-key
  template: |
    {
      "namespace": {{identity.entity.aliases.${auth/kubernetes/@accessor}.metadata.service_account_namespace}}
    }

```
The operator resolves `${auth/kubernetes/@accessor`} to the actual accessor value (e.g. auth_kubernetes_804f1655) before writing the role to Vault.

Note: The Vault role used for authentication must have read and list access to sys/auth for resolution to work. Unresolved placeholders are left as-is.

### DIFF
--- a/api/v1alpha1/identitytokenrole_types.go
+++ b/api/v1alpha1/identitytokenrole_types.go
@@ -148,6 +148,14 @@ func (d *IdentityTokenRole) IsInitialized() bool {
 }
 
 func (d *IdentityTokenRole) PrepareInternalValues(context context.Context, object client.Object) error {
+	if d.Spec.Template == "" {
+		return nil
+	}
+	resolved, err := vaultutils.ResolveAuthAccessors(context, d.Spec.Template)
+	if err != nil {
+		return err
+	}
+	d.Spec.Template = resolved
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


raffaelespazzoli:
can you rebase and resolve the conflicts

Issue #313, opened 2026-04-13
Existing secret values get recreated via RandomSecret even when no rotation is configured and policy is Retain (after #300)


tbalzer:
We observe that after updating from 0.8.38 to 0.8.46 secret values get recomputed in vault now, although this is undesired as they have no rotation configured and the secret value already existed.

We suspect https://github.com/redhat-cop/vault-config-operator/pull/300#issuecomment-4071035532 is the cause for this.

raffaelespazzoli:
this should be fixed in released 0.8.48

PR #315, opened 2026-04-15
fix: preserve existing one-off RandomSecret values on K8s resource re…

raffaelespazzoli:
…create (#313)

PR #300 introduced a regression where one-off RandomSecrets (no refreshPeriod) would overwrite their existing Vault values when the K8s resource was deleted and recreated with kvSecretRetainPolicy: Retain. This happened because the guard that prevented overwriting was removed when adding multi-key support.

Add a preserveExistingKeys parameter to CreateOrMergeKV so that keys already present in Vault are skipped during merge for one-off secrets, while still allowing new keys to be added (multi-key use case). Also fix a duplicate ManageOutcomeWithRequeue call in the reconcile loop.

Made-with: Cursor


sabre1041 (approved):
LGTM

PR #362, opened 2026-08-19
docs: point README at current per-engine docs and HashiCorp Developer…

raffaelespazzoli:
… URLs

Replace stale vaultproject.io links, broken heading fragments, and the missing OperatorHub screenshot so the README lists every supported CRD against live documentation.


sabre1041 (approved):
LGTM

PR #364, opened 2026-08-20
fix: persist OpenShift version annotation and Namespace bundle metadata

raffaelespazzoli:
operator-sdk --overwrite drops custom bundle annotations, so merge them from config/manifests/annotations.yaml after generate. Also declare the Namespace owned CRD description so community-operators static checks stop warning.


sabre1041 (approved):
LGTM

